PR #55, opened 2019-07-20
Fixes for lights, fans, and contact sensors in home assistant

jshridha:
The device payload or lights has been changed and should work for all zwave lights. Lights now will turn on to their previous state if a brightness is not specified.

The fan payload for the GE/Jasco fans is now fixed so that the fan does not always turn on "high" by default. It will turn on to its previous state by default.

### DIFF
--- a/hass/configurations.js
+++ b/hass/configurations.js
@@ -27,8 +27,8 @@ module.exports = {
     type: 'binary_sensor',
     object_id: 'contact',
     discovery_payload: {
-      payload_on: false,
-      payload_off: true,
+      payload_on: true,
+      payload_off: false,
       value_template: '{{ value_json.value }}',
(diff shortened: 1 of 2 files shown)


jshridha:
I added in fixes for contact sensors. The state was being sent in reverse. I have now tested all of my devices and these fixes along with #56 are the issues I currently have.

jshridha:
I've been living with this setup for two days now and it's been working great. No issues so far. This commit is ready for review.

robertsLando:
Sorry Jay I have been really busy this weekend I will try to review this this afternoon or at max tomorrow morning. Thanks for your contribution! :) 

Daniel

robertsLando:
Thanks @jshridha ! :smiley:

PR #62, opened 2019-07-23
[fix] add alarm type, level, and burglar sensors to hass

jshridha:


robertsLando:
Thanks @jshridha ! :)

PR #67, opened 2019-07-24
[fix] Mixed up alarm type and level

jshridha:
Sorry! I mixed these two up and need to send a quick patch.


robertsLando:
Thanks @jshridha

PR #70, opened 2019-07-27
[fix] GE fan speeds not reporting correctly

jshridha:
Fans were not setting their speed correctly anymore. This PR fixes that issue.


robertsLando:
Thanks! :)

PR #85, opened 2019-08-07
Update ControlPanel.vue

alfaromeotango:
Fix spelling in inclusion confirmation
Add confirmation when performing a hard reset


robertsLando:
Thanks @alfaromeotango :)

PR #95, opened 2019-08-12
[fix] Disable z-wave node polling if polling is not set in settings

mattgruter:
Currently, it is not possible to disable node polling once that it has been enabled. This PR fixes that.


mattgruter:
This PR fixes bug #96

robertsLando:
@mattgruter Thanks!

PR #125, opened 2019-09-21
Fix command_on_template for light_dimmer

dchesterton:
The command_on_template should return a value between 0 and 99

### DIFF
--- a/hass/configurations.js
+++ b/hass/configurations.js
@@ -323,7 +323,7 @@ module.exports = {
       state_topic: true,
       state_template: '{{ "off" if value_json.value == 0 else "on" }}',
       command_topic: true,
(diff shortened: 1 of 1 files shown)


robertsLando:
@dchesterton Thanks! :)

jshridha:
@robertsLando This actually reverts the code that was put in place by 771c68e20373f1aebf5c92b83a06b720af75b8f9. I agree that the brightness should return 0-99 for setting a new brightness. However, setting to 255 tells the light to turn on to the previous brightness that was set before the light was turned off. This is why 771c68e20373f1aebf5c92b83a06b720af75b8f9 changed it from 99 to 255. I believe this commit should be reverted, but we can discuss if there's other opinions or if other lights act differently.

PR #135, opened 2019-10-01
Dimmers turn on to previous brightness by default

jshridha:
This reverts the code in #125 so that dimmers can turn on to their previous brightness.


robertsLando:
Thanks @jshridha

PR #154, opened 2019-11-02
added quotes around payload on and off to fix discover in hass

ksuclipse:
Not having the quotes causes HASS to throw errors


ksuclipse:
Did not work as expected

Issue #178, opened 2019-12-24
[bug] 2.0.6-dev crashes 


docbobo:
**Version**

Build/Run method
- [x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

Zwave2Mqtt version: 2.0.6-dev
Openzwave Version: 1.6.962

**Describe the bug**
I just noticed that Z2M was crashing from time to time, trying to re-add a node that is currently marked as dead. No step to reproduce, but the following stacktrace:

> zwave2mqtt    |   z2m:Zwave ValueAdded: 7-134-1-2 Application Version +1ms
> zwave2mqtt    |   z2m:Zwave Notification from node 17: Notification - Node Dead (5) +26ms
> zwave2mqtt    | events.js:116
> zwave2mqtt    |     handler.call(self, arg1);
> zwave2mqtt    |             ^
> zwave2mqtt    |
> zwave2mqtt    | TypeError: Cannot read property 'length' of undefined
> zwave2mqtt    |     at Gateway.publishDiscovery (/snapshot/Zwave2Mqtt/lib/Gateway.js:0:0)
> zwave2mqtt    |     at Gateway.onNodeStatus (/snapshot/Zwave2Mqtt/lib/Gateway.js:0:0)
> zwave2mqtt    |     at emitOne (events.js:116:13)
> zwave2mqtt    |     at ZwaveClient.emit (events.js:211:7)
> zwave2mqtt    |     at ZwaveClient.nodeReady (/snapshot/Zwave2Mqtt/lib/ZwaveClient.js:0:0)
> zwave2mqtt    |     at emitTwo (events.js:126:13)
> zwave2mqtt    |     at OZW.emit (events.js:214:7)

docbobo:
Okay, this might be a dupe of #99. Looking into it, it is cause by a node hat I modified according to your suggestion in #177. It seems that for each component that I set to {}, I will get a callback where hassDevice.values is null. Does that make any sense?

robertsLando:
I have no way to check it today but it could be caused by the change I suggested to you yesterday (set the discoverPayload to {}). Try edit the file nodes.json and remove the hassDevices

Daniel

Issue #99, opened 2019-08-15
[bug] constant crash after startup with ZW095


tagno25:
**Version**

Build/Run method
- [x] Docker
- [ ] PKG
- [x] Manually built (git clone - npm install - npm run build ) 

Zwave2Mqtt version: 2.0.2 / commit a521889
Openzwave Version: 1.4.1

**Describe the bug**
Crash on start after adding Aeon ZW095-A (Home Energy Meter)

**To Reproduce**
1. Add Aeon ZW095-A
2. restart docker container
4. crash after controller becomes "Alive", but before any other node does

**Expected behavior**
No crash

**Additional context**
exact same error with version 2.0.2 or building a docker image using commit a521889
```
  z2m:Zwave Notification from node 16: Notification - NoOperation (2) +31ms
  z2m:Zwave Notification from node 17: Notification - NoOperation (2) +26ms
  z2m:Zwave node 1 ready: AEON Labs - ZW090 Z-Stick Gen5 US (Static PC Controller) +4ms
  z2m:Zwave Notification from node 17: Notification - TimeOut (1) +16s
  z2m:Zwave Notification from node 17: Notification - TimeOut (1) +11s
  z2m:Zwave Notification from node 17: Notification - TimeOut (1) +11s
  z2m:Zwave node 2 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +28ms
  z2m:Zwave node 3 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +26ms
  z2m:Zwave node 4 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +27ms
  z2m:Zwave node 5 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +23ms
  z2m:Zwave node 6 ready: GE (Jasco Products) - 14294 In-Wall Smart Dimmer (Light Dimmer Switch) +32ms
  z2m:Zwave node 7 ready: GE (Jasco Products) - 14287 Fan Control Switch (Fan Switch) +21ms
  z2m:Zwave node 8 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +13ms
  z2m:Zwave node 9 ready: GE (Jasco Products) - 14291 In-Wall Smart Switch (On/Off Power Switch) +12ms
[...]

PR #179, opened 2019-12-25
Fixed crash when components are set to {}

docbobo:
This should fix issue reported in #99 and #178. Might not be perfect, but definitely better than having Z2M crash without any good hint about the reason.


robertsLando:
Thanks @docbobo sorry for the late reply but I was on vacation and I missed some emails

robertsLando:
LGTM

PR #195, opened 2020-01-11
Spelling and consistency (no change in behavior)

chickenandpork:
"charaters" is a spelling error; it may still need commas -- Oxford, of course!  Additionally, slight inconsistency in uppercase/lowercase, and pluralism agreement.

There remains an outstanding oddity on multi instance association description.

Additionally, line 721 should be "nodes' names and locations", but since " ' " is the delimiter on the text string, I can't use it as an apostrophe.  Maybe "node names and locations" remains valid english.


robertsLando:
@chickenandpork Thanks!

PR #200, opened 2020-01-14
Enable sending raw zwave data

derrohrbach:
This patch allows using the api topic to send raw Z-Wave packets. This is extremely useful for devices like the EUROtronic SPIRITZ thermostat, which requires raw messages to use an external temperature source.

Note: This api is only available in very recent versions of OpenZWave.


robertsLando:
LGTM

j9brown:
Is there anything left to do with this pull request?  This function would be very useful to me right about now.  :)

PR #225, opened 2020-01-29
Support dynamic reads of customDevices.json

nervetattoo:
If you use a customDevices.json this will now be re-read on every node discovery instead of being read once when you start the server.
This makes it easier to build out z-wave networks without being forced into network scans due to restarts

Dynamic parsing of customDevices.js is not suported (would either require clearing require cache hack or evaling) but it is still read at startup to continue supporting it.

It also adds `disableHostCheck` for webpack dev-server to enable reaching it over something else than localhost when you set `HOST=0.0.0.0`

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -2,7 +2,9 @@
 /* eslint-disable one-var */
 'use strict'
 
+const fs = require('fs')
 const reqlib = require('app-root-path').require
+const utils = reqlib('/lib/utils.js');
 const EventEmitter = require('events')
 const comandClass = reqlib('/lib/Constants.js').comandClass
 const debug = reqlib('/lib/debug')('Gateway')
@@ -11,25 +13,18 @@ const hassCfg = reqlib('/hass/configurations.js')
 const hassDevices = reqlib('/hass/devices.js')
 const version = reqlib('package.json').version
 
-debug.color = 2
+const CUSTOM_DEVICES = utils.joinPath(true, reqlib('config/app.js').storeDir, 'customDevices')
 
-try {
-  const storeDir = reqlib('config/app.js').storeDir
-  const customDevices = reqlib(storeDir + '/customDevices')
-  let i = 0
-  for (const d in customDevices) {
-    hassDevices[d] = customDevices[d]
-    i++
+// Continue supporting customDevices.js for now
+if (fs.existsSync(CUSTOM_DEVICES + '.js')) {
+  const devices = reqlib(CUSTOM_DEVICES + '.js')
+  for (const key in devices) {
+    hassDevices[key] = devices[key]
   }
-
-  debug('Loaded', i, 'custom hass devices configurations')
-} catch (error) {
-  if (error.message.indexOf('Cannot find module') >= 0) {
-    debug('No custom hass devices file found')
-  } else {
-    debug('Error while loading store/customDevices', error.message)
-  }
+  debug('Loaded', Object.keys(devices).length, 'custom js hass devices configurations')
 }
+
+debug.color = 2
 
 const NODE_PREFIX = 'nodeID_'
 // const GW_TYPES = ['valueID', 'named', 'manual']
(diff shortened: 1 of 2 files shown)


lib/Gateway.js
```diff
@@ -11,26 +13,46 @@ const hassCfg = reqlib('/hass/configurations.js')
...
+        ...hassDevices,
+        ...devices
+      }
+    } else {
+      debug('No custom hass devices file found')
+    }
+  } catch (error) {
     debug('Error while loading store/customDevices', error.message)
   }
+  return { ...hassDevices }
```
robertsLando: Why not return `hassDevices` here?

lib/Gateway.js
```diff
@@ -11,26 +13,46 @@ const hassCfg = reqlib('/hass/configurations.js')
...
+}
 
-  debug('Loaded', i, 'custom hass devices configurations')
-} catch (error) {
-  if (error.message.indexOf('Cannot find module') >= 0) {
-    debug('No custom hass devices file found')
-  } else {
+debug.color = 2
+
+debug(`storeDir: ${STORE_DIR}`)
```
robertsLando: Could you remove this?

lib/Gateway.js
```diff
@@ -11,26 +13,46 @@ const hassCfg = reqlib('/hass/configurations.js')
...
 
-try {
-  const storeDir = reqlib('config/app.js').storeDir
-  const customDevices = reqlib(storeDir + '/customDevices')
-  let i = 0
-  for (const d in customDevices) {
-    hassDevices[d] = customDevices[d]
-    i++
+// Continue supporting customDevices.js for now
+if (fs.existsSync(path.join(STORE_DIR, 'customDevices.js'))) {
```
robertsLando: I think we could remove this and keep just the `loadDevices` method. Add a var  `this.customDevicesLength` in Gateway constructor and set it to `0`. Every time customDevices is read in loadDevices metohd check for how many keys it has and if different than the previous `customDevicesLength` add a debug:

` debug('Loaded', Object.keys(devices).length, 'custom js hass devices configurations')`

robertsLando:
Thanks for this PR, would you mind fix those little points?

lib/Gateway.js
nervetattoo: loadDevices doesn't read the js file because it would be cached by require and it seemed clearer to make them distinct. At first I thought of deprecating the js file, but that would be breaking for someone so I kept the startup read of it.
I guess we can track the change in count yes.

nervetattoo:
Changed and amended to last commit

lib/Gateway.js
```diff
@@ -199,18 +197,25 @@ function onValueChanged (valueId, node, changed) {
...
+    const deviceKeys = Object.keys(allDevices)
+    if (deviceKeys.length !== this.customDevicesLength) {
+      this.customDevicesLength = deviceKeys.length
+      debug(`Loaded ${this.customDevicesLength} custom hass device configurations: ${deviceKeys.join(', ')}`)
     }
+    const nodeDevices = allDevices[node.device_id] || []
+    // const configs = Object.keys(allDevices).join(', ')
+    // debug(`node ${node.node_id} deviceId ${node.device_id}`)
+    // debug(`configs ${configs}`)
+    // debug(`found ${nodeDevices.length} configurations`)
```
robertsLando: Could you remove those unused comments?

lib/Gateway.js
```diff
@@ -11,26 +13,20 @@ const hassCfg = reqlib('/hass/configurations.js')
...
-try {
-  const storeDir = reqlib('config/app.js').storeDir
-  const customDevices = reqlib(storeDir + '/customDevices')
-  let i = 0
-  for (const d in customDevices) {
-    hassDevices[d] = customDevices[d]
-    i++
-  }
+const STORE_DIR = reqlib('config/app.js').storeDir
+const CUSTOM_DEVICES = 'customDevices.json'
```
robertsLando: Import `lib/utils` and use` utils.joinPath` because it keep sure the path is correct when using `pkg` files.
Remove STORE_DIR var and use just the:

```js
const CUSTOM_DEVICES = utils.joinPath(true, reqlib('config/app.js').storeDir, 'customDevices')
```

And add `js|json` extensions on the methods

robertsLando:
LGTM. Thanks @nervetattoo

PR #236, opened 2020-02-04
set sensor_home_security to index 7, sensor_access_control to 6

chilicheech:
Not all alarm command class indeces below 256 are burglar according to https://github.com/OpenZWave/open-zwave/blob/4478eea26b0e1a29184df0515a8034757258ff88/cpp/src/ValueIDIndexesDefines.def#L1068

This PR sets sensor_home_security to index 7 and sensor_access_control to index 6.

Cheers,

### DIFF
--- a/hass/configurations.js
+++ b/hass/configurations.js
@@ -132,25 +132,9 @@ module.exports = {
       value_template: '{{ value_json.value }}'
     }
   },
-  'sensor_alarm_level': {
+  'sensor_alarm': {
     type: 'sensor',
-    object_id: 'alarm_level',
-    discovery_payload: {
-      icon: 'mdi:alarm-light',
-      value_template: '{{ value_json.value }}'
-    }
-  },
-  'sensor_alarm_type': {
-    type: 'sensor',
-    object_id: 'alarm_type',
-    discovery_payload: {
-      icon: 'mdi:alarm-light',
-      value_template: '{{ value_json.value }}'
-    }
-  },
-  'sensor_alarm_burglar': {
-    type: 'sensor',
-    object_id: 'alarm_burglar',
+    object_id: 'alarm',
     discovery_payload: {
       icon: 'mdi:alarm-light',
       value_template: '{{ value_json.value }}'
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -856,21 +856,40 @@ Gateway.prototype.discoverValue = function (node, valueId) {
       case 'sensor_alarm':
         cfg = copy(hassCfg.binary_sensor_alarm)
         break
+
       case 'alarm':
-
-        if (valueId.index < 256) {
-          cfg = 'sensor_alarm_burglar'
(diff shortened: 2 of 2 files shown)


robertsLando:
@chilicheech Those configurations doesn't exists in hass/devices.js

chilicheech:
@robertsLando I made modifications to add all devices below 256. There's an issue though. There is a `sensor_water` in the *Alarm/Notification* command class that is duplicated by the *SensorMultiLevel* command class. We need to figure out a way to namespace them.
Look at `hass/configurations.js:` lines 171 and 502. Also look at:
https://github.com/OpenZWave/open-zwave/blob/master/config/Localization.xml#L67
The *Alarm/Notification* water sensor is the only one named *water*. The *SensorMultiLevel* command class has some water sensors but they're named *water_....*. None of them are named *water*. We could rename that one sensor in `hass/configurations.js` and not have to worry about namespacing.
What do you suggest?

chilicheech:
OK, I created new configurations and relabeled the original `sensor_water` according to the localization document.

robertsLando:
Did you made a PR for this? 

@chilicheech I suggest to don't rename existing configurations, please use `sensor_alarm_<alarmTypeHere>` instead.

robertsLando:
Also I suggest to keep just keep generic `sensor_alarm` configuration like this:

```js
'sensor_alarm': {
    type: 'sensor',
    object_id: 'alarm',
    discovery_payload: {
      icon: 'mdi:alarm-light',
      value_template: '{{ value_json.value }}'
    }
  },
```

And then use your map here to set the correct object id:

```js
let valueIndexConfigMap = {         
        1: 'smoke',
         2: 'carbon_monoxide',
         3: 'carbon_dioxide',
         4: 'heat',
         5: 'water',
         6: 'access_control',
         7: 'home_security',
         8: 'power_management',
         9: 'system',
         10: 'emergency',
         11: 'clock',
         12: 'appliance',
         13: 'home_health',
         14: 'siren',
         15: 'water_valve',
         16: 'weather',
         17: 'irrigation',
         18: 'gas',
         512: 'type',
         513: 'level'
       }

let alarm_type = valueIndexConfigMap[valueId.index]	   

if(alarm_type) {
cfg = copy(hassCfg[cfg])
cfg.object_id += alarm_type
}   
```

chilicheech:
That's a good idea. I've updated the changes. I consolidated `sensor_alarm_burglar`, `sensor_alarm_type`, and `sensor_alarm_level` into `sensor_alarm` and, like you suggested, set the object id based on the alarm type. That way there's much less code duplication.

robertsLando:
LGTM. Thanks @chilicheech

robertsLando:
@chilicheech Would you mind to check also sensors multilevel?

chilicheech:
@robertsLando sounds good. I'll take a look at the sensor multilevel.

Cheers,

PR #252, opened 2020-02-10
refactor: sensor multilevel constants

chilicheech:
This refactors how we determine multilevel sensor types to be similar to how we do alarm sensors. It includes the names for the object ids.

@robertsLando there are some cases where the object id has the same name as the sensor type, so I kept it blank to avoid names like `direction_direction`. Let me know what you think.

Also, some sensors look like they belong to other types, but I didn't want to move them around just yet. For example, sensor 57 is under *sensor_water*, but it's a water pressure sensor, and maybe it should be under *sensor_pressure* instead..

Another thing to consider is in the future we may want to populate not just object ids but also units of measurement based on the constants.

It would be really great if instead of redefining things in the Constants.js if we could just read in the open-zwave xml files that it's based on and parse them to the correct attributes, i.e.: config/Localization.xml and config/SensorMultiLevelCCTypes.xml

Cheers,

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -1,6 +1,6 @@
 module.exports = {
   meterType (index) { // https://github.com/OpenZWave/open-zwave/blob/0d94c9427bbd19e47457578bccc60b16c6679b49/cpp/src/command_classes/Meter.cpp#L74
-    var cfg = null
+    let cfg = null
     if (index >= 16 && index < 32) {
       cfg = 'sensor_gas'
     } else if (index < 48) {
@@ -13,7 +13,7 @@ module.exports = {
   alarmType (index) {
     // https://github.com/OpenZWave/open-zwave/blob/4478eea26b0e1a29184df0515a8034757258ff88/cpp/src/ValueIDIndexesDefines.def#L1068
     // https://github.com/OpenZWave/open-zwave/blob/05a096f75dddd27e3f8dc6af2afdb3cad3b4ebaa/config/Localization.xml#L7
-    var alarmMap = {
+    let alarmMap = {
       1: 'smoke',
       2: 'carbon_monoxide',
       3: 'carbon_dioxide',
@@ -32,6 +32,8 @@ module.exports = {
       16: 'weather',
       17: 'irrigation',
       18: 'gas',
+      256: 'previous_event_cleared',
+      257: 'alert_location',
       512: 'type',
       513: 'level'
     }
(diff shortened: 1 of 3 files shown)


lib/Constants.js
```diff
@@ -32,92 +32,125 @@ module.exports = {
...
     }
+
+    let sensorType = {
+      sensor: 'sensor_generic',
+      objectId: 'unknown_' + index
+    }
+
+    for (let sensor in sensorMap) {
+      let objectId = sensorMap[sensor][index]
+      if (objectId === null) continue // objectId can be an empty string
```
robertsLando: This should be `objectId === undefined`

lib/Constants.js
```diff
@@ -32,92 +32,125 @@ module.exports = {
...
-        return 'sensor_time'
-      case 44:
-      case 45:
-      case 50:
-      case 60:
-      case 69:
-        return 'sensor_heart'
-      default:
-        return null
+    let sensorMap = {
```
robertsLando: There are some more sensors in the list, also I suggest to add another map with sensors icons so we can keep only one sensor generic configuration and edit the object_id and the discovery payload mdi icon

chilicheech:
I added the missing sensors and changed null to undefined.
As for the other map for the icons, I think it would be better to do that in a separate PR. We could also do something similar for the units of measurement.

The more I look at this the more I think we should just read in the config/NotificationCCTypes.xml and config/SensorMultiLevelCCTypes.xml files and parse them into objects we can use. They have `id` and `name` that we can use. Name can be changed to lower case and special characters changed to `_`. And for the sensor multilevel we could even use the first unit of measurement in the list.

This way we don't have to keep our own maps, and just use what's supplied by OZW. The only map we would need is for the icons.

robertsLando:
@chilicheech I understand what you mean but I prefer to keep my own map for now, the best thing would be getting them from OZW directly instead of reading xml files but that requires a change in openzwave-shared library

robertsLando:
> As for the other map for the icons, I think it would be better to do that in a separate PR. We could also do something similar for the units of measurement.

I prefer to merge this once we clean up all sensors, there is still lot confusion in configuration.js file too many repeating configurations. About units of measurements them are overrided by valueId units so they can all be removed from configurations

Thanks for your help @chilicheech

chilicheech:
Thanks. I saw the modifications you made with the `props` field. Looks good.

Cheers,

robertsLando:
Thanks to you! :)

Daniel

PR #258, opened 2020-02-17
Fix electricity meter named as water meter

ovenystas:
Electricity kwh meter got classified and named as a water meter instead as an electricity meter as it is supposed to be. sensor_water_kwh_meter instead of sensor_electricity_kwh_meter.


robertsLando:
@ovenystas Thanks

PR #261, opened 2020-02-18
Add "remove failed node" command

billiaz:
Adding one more node command - removeFailedNode 

- this command can be used after a node is Dead or "Has node failed" been executed against node.
- This command can delete/remove permanently a node from the controller


robertsLando:
LGTM. Thanks @billiaz !

PR #267, opened 2020-02-20
Make Nodes in Groups/Associations more easy to identify 

billiaz:
When using Node Names instead of Node IDs. I have had straggle to create associations for similar Items. as they have had the same title.

I added a new pattern which creates the {location}_{name} instead of just {name} as label. This makes it easy to find which device you try to update the associations.

There pattern is as:

| Name | Location | final node label| 
|-------|----------|----------|
| TRV | Bedroom | Bedroom_TRV |
| TRV | Masterbedroom | Masterbedroom_TRV|
|SWITCH | '' | _SWITCH|

when the Location is empty you receive an leading Underscore. This can be corrected to hide it with extra condition. Or use space instead of underscore when joining the values.

I find it more easy to know when underscore exists what info you receive.

Screenshot form my compiled container:

Screenshot 2020-02-20 at 14 55 13


src/components/ControlPanel.vue
```diff
@@ -1054,7 +1054,7 @@ export default {
       this.setName(n)
     },
     setName (n) {
-      n._name = n.name || 'NodeID_' + n.node_id
+      n._name = [n.loc, n.name].join('_') || 'NodeID_' + n.node_id
```
robertsLando: What if node loc doesn't exists?

billiaz: if n.loc is empty, you get an _<name> see my screenshot which is _test

robertsLando: @billiaz I would prefer more user friendy names (some users may have underscore in their node names so I suggest: `n.name (n.location)` if location doesn't exist just use `n.name`

billiaz: n.loc and n.location returns the same?

robertsLando: @billiaz Maybe I explained it wrong, I mean this `n.name (n.location)` is the result I want. Translated in code it becomes: `n.name + (n.loc ? ' (' + n.loc + ')' : ''`

robertsLando:
fe9511e0da043b29e9a3561b43b4acf2f1575f5a

PR #269, opened 2020-02-21
Central Scene Sensor 

billiaz:
I have a NodOn remote (but all remotes has similar way of working).

Sometimes you have scenes which they do change state. This patch creates sensors which will allow to monitor and automate upon these in HASS.

One concern, can I use the value_json.label as name in the scene, instead of the id? if so.. what I need to change here?

example of the result in HASS
Screenshot 2020-02-21 at 17 36 52


robertsLando:
LGTM, Thanks!

PR #270, opened 2020-02-22
make rgb_dimmer have state in caps and use RGB

### LINKED ISSUE #68: [question] Light discovered in Hassio without rgb
I am migrating from existing Zwave integration in Hassio. Zipato Bulb 2 was in Hassio(installed ZWAVE2MQTT add-on) discovered as `light.nodeid_24_dimmer`, without RGB option. In ZWAVE2MQTT add-on the light Type is Light Dimmer Switch-same as Fibaro Dimmer 2. At Home Assistant - Devices from the light node, there are `light_dim` and `light_rgb` devices, but I was not able to push `light_rgb`-get colors to Hassio. In add-on I can change colors, etc.. How can I change this? Dimmer and switches had been discovered okay.

jwalker343:
I think this fixes #68 , @robertsLando, but I'm not sure if I should change this here for all rgb_dimmers, or if we should break out the two specific models mentioned in the issue in `hass/devices.js`
- Zipato Bulb 2
- Aeon Labs ZW098

Thoughts?

### DIFF
--- a/hass/configurations.js
+++ b/hass/configurations.js
@@ -139,7 +139,7 @@ module.exports = {
     discovery_payload: {
       state_topic: true,
       command_topic: true,
-      rgb_command_template: '{{ \'#%02x%02x%02x\' | format(blue, green, red)}}',
+      rgb_command_template: '{{ \'#%02x%02x%02x\' | format(red, green, blue)}}',
       rgb_command_topic: true,
       rgb_state_topic: true,
       rgb_value_template: '{{ value_json.value[1:3] | int(0, 16) }},{{ value_json.value[3:5] | int(0, 16) }},{{ value_json.value[5:7] | int(0, 16) }}'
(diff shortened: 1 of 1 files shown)


robertsLando:
@jwalker343 Did you tested this? Wht you changed `off|on` to `ON|OFF` ?

tekmaven:
@robertsLando upgrading to the latest build broke my lights state.  The brightness settings would work properly, but the light would never show as "on".  

Reverting to the commit prior to this merge, c0500d277fcaafdbce307857f7324da132b04b23, fixes my issue.

robertsLando:
@tekmaven Are you using latest hass version? Because as you can see on readme now I try to keep ozw compatible with latest hass version. This will break old hass instances for sure

hass/configurations.js
```diff
@@ -154,9 +154,9 @@ module.exports = {
       brightness_state_topic: true,
       brightness_command_topic: true,
       on_command_type: 'first',
-      state_value_template: '{{ "off" if value_json.value == 0 else "on" }}',
+      state_value_template: '{{ "OFF" if value_json.value == 0 else "ON" }}',
```
balloob: This is incorrect. 

https://github.com/home-assistant/core/blame/dev/homeassistant/components/mqtt/light/schema_template.py#L223-L226

For 3 years this has been `on` and `off`.

chilicheech: I submitted a fix for it with #289, just waiting for the merge :)

jwalker343: Looking at those lines in the actual code, @robertsLando, this is fuzzier than expected, it appears inconsistent in the code, and the docs may not be clear/accurate. 

This worked for me in testing, but could be because my RGB Lights specifically are expecting payload_on/off to be all caps. I do not have any normal on/off lights.

I am out of town, and down for whatever makes the most sense. IMO, If this is breaking a lot of peoples stuff, we should revert the change and then open an issue with home-assistant core to investigate further.

robertsLando: I will submit a patch release in next ours

PR #272, opened 2020-02-25
Fix temperature sensor's unit of measurement for HomeAssistsant

adrum:
This fixes the unit of measurement for temperature sensors when being used in HomeAssistant's auto discovery feature. My sensors were throwing errors when being used inside of HomeAssistant. 

https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/const.py#L343
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/util/temperature.py#L28

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -943,6 +943,9 @@ Gateway.prototype.discoverValue = function (node, valueId) {
 
         if (valueId.units) {
           cfg.discovery_payload.unit_of_measurement = valueId.units
+          if (['C', 'F'].indexOf(valueId.units.toUpperCase()) > -1) {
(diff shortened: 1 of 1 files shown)


robertsLando:
@adrum Thanks for your pr but I can't accept it because after long discussion with openzwave author he has highly recommended to not use string compare inside code, even for units, because starting from ozw 1.6 there is multi language support and those will fail. I suggest you to fix the configurations files of your devices in ozw database

Fishwaldo:
There is not a config file variable here. There is a ValueID exposed for each sensor  (>255) that allows you to select which “unit” the device should send (C/F). You could use that to find the available scales and which one is selected. 

That being said - there are devices out there that ignore it and have a configuration option you need to change to change the scale.

adrum:
Ok no problem. Thank you for your consideration. I was originally going to go down that path, but thought this approach might worthwhile. 

Thank you providing this great software.

PR #276, opened 2020-02-28
Change the values published towards HASS Discovery

billiaz:
In order to make devices identifiable in Home Assistant, I suggest to change the payload to the following attributes:

1. Model: Change the ProductID (HEX) to Product Name
2. Name: Change the Product Name to nodeName

This way devices will be named in a different way in HASS, and will be easy to clean up the Database

End result:

Screenshot 2020-02-28 at 12 25 19


lib/Gateway.js
```diff
@@ -425,8 +425,8 @@ function deviceInfo (node, nodeName) {
   return {
     identifiers: ['zwave2mqtt_' + nodeName],
     manufacturer: node.manufacturer,
-    model: node.productid,
-    name: node.product,
+    model: node.product,
```
robertsLando:
```suggestion
    model: node.product + ' (' + node.productid + ')',
```

PR #302, opened 2020-03-06
use fan_mode_map instead of mode_map for thermostat_fan_mode command class

chilicheech:
With the current behavior for thermostat/climate devices values received from HASS for fan modes are translated using the mode_map. However, the mode_map only maps modes for the hvac mode, not the fan mode.

This change makes it so the fan mode command class will map its values against the fan_mode_map instead of the mode_map.

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -489,7 +489,7 @@ Gateway.prototype.parsePayload = function (payload, valueId, valueConf) {
 
     // Hass payload parsing
     if (this.discovered[valueId.value_id]) {
-      // parse payload for switchs
+      // parse payload for switches
       if ((valueId.type === 'bool' || this.discovered[valueId.value_id].object_id === 'rgb_dimmer') && typeof payload === 'string') {
         if (/\btrue\b|\bon\b|\block\b/gi.test(payload)) payload = true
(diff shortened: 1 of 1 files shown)


lib/Gateway.js
```diff
@@ -502,15 +502,18 @@ Gateway.prototype.parsePayload = function (payload, valueId, valueConf) {
 
       // map modes coming from hass
       if (valueId.type === 'list' && valueId.values.indexOf(payload) < 0) {
-        var hassDevice = this.discovered[valueId.value_id]
-        payload = hassDevice.mode_map ? hassDevice.mode_map[payload] : payload
+        let hassDevice = this.discovered[valueId.value_id]
+        // for thermostat_fan_mode command class use the fan_mode_map
+        if (valueId.class_id === 0x44 && hassDevice.fan_mode_map) payload = hassDevice.fan_mode_map[payload]
```
robertsLando: You should add a check that `hassDevice` is not defined

chilicheech: Good idea. I added a check. See if that's good or if we need to specifically check for undefined.

robertsLando:
LGTM

PR #305, opened 2020-03-06
render the action_topic and reformat the code a little

### LINKED ISSUE #304: [bug] climate action_topic not being rendered
**Version**

Build/Run method
- [x] Docker
- [ ] PKG

Zwave2Mqtt version: docker 3.0.1
Openzwave Version: docker 3.0.1

**Describe the bug**
According to the documentation [here|https://github.com/OpenZWave/Zwave2Mqtt#thermostats-with-fans] I've set the `action_topic` to `66-1-0`, which is the correct value for my thermostat. However, the rendered hass discovery payload is:
```
{
  "type": "climate",
  "object_id": "thermostat",
  "values": [
    "49-1-1",
    "64-1-0",
    "66-1-0",
    "67-1-1",
    "67-1-2",
    "68-1-0"
  ],
  "mode_map": {
    "off": "Off",
    "heat": "Heat",
    "cool": "Cool"
  },
  "fan_mode_map": {
    "auto": "Auto Low",
    "on": "On Low"
  },
  "setpoint_topic": {
    "Heat": "67-1-1",
    "Cool": "67-1-2"
  },
  "default_setpoint": "67-1-1",
  "discovery_payload": {
    "min_temp": 50,
    "max_temp": 85,
    "modes": [
      "off",
      "heat",
      "cool"
    ],
    "fan_modes": [
      "auto",
      "on"
    ],
    "action_topic": "66-1-0",
    "action_template": "{{ value_json.value }}",
    "current_temperature_topic": "z2m/31/49/1/1",
    "current_temperature_template": "{{ value_json.value }}",
    "fan_mode_state_topic": "z2m/31/68/1/0",
    "fan_mode_command_topic": "z2m/31/68/1/0/set",
    "mode_state_topic": "z2m/31/64/1/0",
    "mode_command_topic": "z2m/31/64/1/0/set",
    "temperature_state_template": "{{ value_json.value }}",
    "temperature_command_topic": "z2m/31/67/1/1/set",
    "mode_state_template": "{% set values = {\"Off\":\"off\",\"Heat\":\"heat\",\"Cool\":\"cool\"} %} {{ values[value_json.value] if value_json.value in values.keys() else 'off' }}",
    "temperature_state_topic": "z2m/31/67/1/1",
    "fan_mode_state_template": "{% set values = {\"Auto Low\":\"auto\",\"On Low\":\"on\"} %} {{ values[value_json.value] if value_json.value in values.keys() else 'auto' }}",
    "device": {
      "identifiers": [
        "zwave2mqtt_0xc5b5c9e5_node31"
      ],
      "manufacturer": "2GIG Technologies",
      "model": "CT32 Thermostat (0x0100)",
      "name": "thermostat",
      "sw_version": "10.00"
    },
    "name": "thermostat_thermostat",
    "unique_id": "zwave2mqtt_0xc5b5c9e5_Node31_thermostat"
  },
  "discoveryTopic": "climate/thermostat/thermostat/config",
  "persistent": false,
  "id": "climate_thermostat"
}
```
As you can see, the `action_topic` has a value of `66-1-0`, which is what I entered, instead of the value it should render: `z2m/31/66/1/0`.

**To Reproduce**

**Expected behavior**
The climate/thermostat's `action_topic` should be rendered from an endpoint id to the correct topic path. In other words, `66-1-0` should be rendered to something like `z2m/31/66/1/0`.

**Additional context**
My current workaround is to enter the actual path instead of the ID.

chilicheech:
This is to fix #304

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -716,11 +716,11 @@ Gateway.prototype.rediscoverAll = function () {
 }
 
 Gateway.prototype.discoverDevice = function (node, hassDevice) {
-  var hassID = hassDevice ? hassDevice.type + '_' + hassDevice.object_id : null
+  let hassID = hassDevice ? hassDevice.type + '_' + hassDevice.object_id : null
 
   try {
     if (hassID && !node.hassDevices[hassID]) { // discover the device
-      var payload
+      let payload
 
       // copy the configuration without edit the original object
       hassDevice = JSON.parse(JSON.stringify(hassDevice))
@@ -728,8 +728,8 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
       if (hassDevice.type === 'climate') {
         payload = hassDevice.discovery_payload
 
-        var mode = node.values[payload.mode_state_topic]
-        var setId
+        let mode = node.values[payload.mode_state_topic]
+        let setId
 
         if (mode !== undefined) {
           setId = hassDevice.setpoint_topic && hassDevice.setpoint_topic[mode.value] ? hassDevice.setpoint_topic[mode.value] : hassDevice.default_setpoint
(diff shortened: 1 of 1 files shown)


chilicheech:
Let me know if the code formatting is ok. Also, I'm not sure we need `if (variable !== undefined)` versus just `if (variable)` since it will always be truthy when it's not undefined. It also weeds out empty strings.

robertsLando:
> I'm not sure we need if (variable !== undefined) versus just if (variable)

Yep it is not needed but I started using that because it is more clear then `if (var)` and sometimes it could prevent errors when for example comparing number values and if value is `0` it will return false and that is not what you want in such cases. It is just a more concise way to make comparison, if the value is not in the map it should be undefined so I use undefined

PR #326, opened 2020-03-25
default precision to 0.1 on hass discovery climate devices

chilicheech:
Have a more precise reading as a default seems to do no harm and will give better reading for those who have thermostats that report temperature with a more granular precision than 1.0 degree.

Users can still override the precision by setting it in the customDevices.json or the devices.js. If no precision is set then the default is used.


robertsLando:
@chilicheech Thanks, LGTM

PR #347, opened 2020-04-05
Add HRT4-ZW device from Hortmann.

cyclops1982:
This should add support fro the Hortmann HRT4-ZW device. 

I've tested this briefly after running it with the customDevices for a while. I copied the values from there.


robertsLando:
@cyclops1982 Thanks!

PR #350, opened 2020-04-07
add 2gig ct32 thermostat

chilicheech:
add 2gig ct32 thermostat, fix lint issues, and update readme thermostat example

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ Gateway settings:
       - `index`: the numerical index of the value
 
   2. **Named Topics**: *Automatically configured*. **DEPRECATED** After a discussion with Openzwave author lib we discourage users to use this configuration as we cannot ensure that value labels will be the same, they could change in future versions (and also they depends on localization added in OZW 1.6). You can find more info [HERE](https://github.com/OpenZWave/Zwave2Mqtt/issues/22)
-  
+
       The topic where zwave values are published will be:
 
       `<mqtt_prefix>/<?node_location>/<node_name>/<class_name>/<?instance>/<value_label>`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -457,24 +457,24 @@ So in decimal format will become: `411-514-3`. This is the device id of `Heatit
 
 #### Thermostats
 
-```js
+```yaml
```
robertsLando, line 460: That's not yaml, it' JSON

README.md
```diff
@@ -497,24 +497,24 @@ Thermostats are most complex components to create, in this device example the se
 
 #### Fans
 
-```js
+```yaml
```
robertsLando, line 500: same

README.md
```diff
@@ -533,18 +533,17 @@ Thermostats are most complex components to create, in this device example the se
 
 The main template is like the thermostat template. The things to add are:
 
-```js
+```yaml
```
robertsLando, line 536: same

robertsLando:
Everything is good here just change yaml with json and we can merge this. Thanks @chilicheech

README.md
chilicheech: I know it's not `yaml`, but since `json` doesn't support comments and there's a comment in there, and  `json` is valid `yaml` I figured I would set it to `yaml` just so the syntax highlighting works properly with IDEs. I can change it to `json` and move the comments to outside the code block.

references:
https://stackoverflow.com/questions/21584985/what-valid-json-files-are-not-valid-yaml-1-1-files
https://en.wikipedia.org/wiki/YAML#Comparison_with_JSON

robertsLando:
Yes you right that’s why I was using js as formatter (not json Sorry) 

Support me at:
    Github sponsors

README.md
```diff
@@ -533,61 +535,61 @@ Thermostats are most complex components to create, in this device example the se
...
-        "values": [
-            "64-1-0",
-            "49-1-1",
-            "67-1-1",
-            "67-1-2",
-            "66-1-0", // <-- add fan values
-            "68-1-0", // <-- add fan values
-            "69-1-0"  // <-- add fan values
+##### GoControl GC-TBZ48 (Linear Nortek Security Control LLC):
+```json
```
robertsLando, line 539:
```suggestion
```js
```

README.md
```diff
@@ -497,24 +498,25 @@ Thermostats are most complex components to create, in this device example the se
...
-      speeds: ["off", "low", "medium", "high"],
-      payload_low_speed: 24,
-      payload_medium_speed: 50,
-      payload_high_speed: 99,
-      payload_off: 0,
-      payload_on: 99,
-      state_value_template: "{% if (value_json.value | int) == 0 %} 0 {% else %} 99 {% endif %}",
-      speed_value_template: "{% if (value_json.value | int) == 25 %}  24  {% elif (value_json.value | int) == 51 %} 50 {% elif (value_json.value | int) == 99 %} 99 {% else %}  0  {% endif %}"
+##### GE 1724 Dimmer:
+```json
```
robertsLando, line 502:
```suggestion
```js
```

README.md
```diff
@@ -457,24 +457,25 @@ So in decimal format will become: `411-514-3`. This is the device id of `Heatit
...
-      max_temp: 30,
-      modes: ['off', 'heat', 'cool'],
-      mode_state_topic: '64-1-0',
-      mode_command_topic: true,
-      current_temperature_topic: '49-1-1',
-      current_temperature_template: '{{ value_json.value }}',
-      temperature_state_template: '{{ value_json.value }}',
-      temperature_command_topic: true
+##### Heatit Thermostat TF 021 (ThermoFloor AS):
+```json
```
robertsLando, line 461:
```suggestion
```js
```

chilicheech:
Ok, so the reason I changed it to json is because of `customDevices.json`. It seems simpler to use that instead of `customDevices.js`, especially because it support auto reloading.

I was looking at the rendered output in GitHub and I think the syntax highlighting looks better with yaml instead of json. With yaml the keys have a different color than the values. With json it's all the same color. And since json is valid yaml it made sense to me. Folks reading the readme won't know that it says yaml, it just helps the application render the visualization and checks correctly.

Also, by switching to `js` my IDE still complains, whereas with `yaml` it doesn't:
js:
g-js
yaml:
g-yaml

I changed it back to `js`. If you want I can change the example back to js from json too.

Cheers,

robertsLando:
Thanks @chilicheech !

PR #363, opened 2020-04-19
Update devices.js for new cover type

Lopton:
Added Graber/Bali/Spring Fashion Covers to devices.js


robertsLando:
@Lopton  Thanks!

PR #364, opened 2020-04-20
allow a zwave plugin to be defined exposing the zwave client to external js

### LINKED ISSUE #359: [feat] expose the zwave client
i've got my own stuff that talks to the zwave interface, (I'm migrating away from that and will just use your API eventually, but until then...)
https://github.com/chrisns/iot-zwave/blob/master/monkey-patch.patch#L22
At the minute I'm doing some gross patching, what would be nicer is to be able to specify something in the settings to include a plugin arch of some sort.

thoughts?
happy to write the code and pr if we can agree the style
similar for exposing the mqtt client would be good, though i've not looked at that side of the gateway just yet

chrisns:
As discussed in and closes #359 [[feat] expose the zwave client](/OpenZWave/Zwave2Mqtt/issues/359)

First time i've touched anything to do with vue, so no idea if this is the right/best way to do it?

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -157,7 +157,12 @@ async function init (cfg, socket) {
   } else {
     this.client = CLIENT = new OpenZWave(options)
     if (cfg.plugin) {
-      require(cfg.plugin)(this)
+      try {
+        require(cfg.plugin)(this)
+      }
+      catch (error) {
+        debug(`Error while loading ${cfg.plugin} plugin`, error.message)
+      }
     }
   }
 
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -156,6 +156,9 @@ async function init (cfg, socket) {
     this.client.updateOptions(options)
   } else {
     this.client = CLIENT = new OpenZWave(options)
+    if (cfg.plugin) {
+      require(cfg.plugin)(this)
+    }
   }
 
   this.nodes = []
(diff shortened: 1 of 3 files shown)


src/components/Settings.vue
```diff
@@ -125,6 +125,7 @@
                         hint="Path to devices library DB. If not set the default path will be used based on your OS"
                       ></v-text-field>
                     </v-flex>
+                    <input type="hidden" :value="zwave.plugin">
```
robertsLando, line 128: This is not necessary you can remove it :)

lib/ZwaveClient.js
```diff
@@ -156,6 +156,9 @@ async function init (cfg, socket) {
     this.client.updateOptions(options)
   } else {
     this.client = CLIENT = new OpenZWave(options)
+    if (cfg.plugin) {
```
robertsLando, line 159: What about also make a check the plugin path exists? `fs.existSync`?

src/components/Settings.vue
chrisns: without it, hitting save in the settings removed it from the `settings.json`

robertsLando: Oh ok, no problem so. Keep it :)

lib/ZwaveClient.js
robertsLando: @chrisns What about this?

chrisns: agree, will push a fix later today

chrisns: put it in a try/catch instead of your recommendation since this allows normal manipulation of the paths that require uses (which might come in handy when we write tests) and felt like a better practice.

robertsLando: It's ok for me :)

robertsLando:
LGTM

PR #367, opened 2020-04-22
feat: arbitrary zwave config

### LINKED ISSUE #358: [feat] arbitrary zwave config
It would be great to be able to specify arbitrary zwave config, at the minute i'm having to patch the `lib/ZwaveClient.js` (ignore the other things in the patch I'll raise a seperate issue to discuss that

https://github.com/chrisns/iot-zwave/blob/master/monkey-patch.patch#L9-L14

Something in the `settings.json` would be great! thanks

if you agree with the principal I'm more than happy to write the code

chrisns:
As discussed in and closes #358 [[feat] arbitrary zwave config](/OpenZWave/Zwave2Mqtt/issues/358)

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -150,6 +150,10 @@ async function init (cfg, socket) {
   if (cfg.assumeAwake) {
     options.AssumeAwake = cfg.assumeAwake
   }
+  
+  if (cfg.options) {
+    options = {...options, ...cfg.options}
+  }
 
   if (CLIENT) {
     this.client = CLIENT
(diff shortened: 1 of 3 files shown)


lib/ZwaveClient.js
```diff
@@ -150,6 +150,10 @@ async function init (cfg, socket) {
   if (cfg.assumeAwake) {
     options.AssumeAwake = cfg.assumeAwake
   }
+  
+  if (cfg.options) {
+    options = {...options, ...cfg.options}
```
robertsLando, line 155: Maybe using `Object.assign` here is more clear ?

`Object.assign(options, cfg.options)`

robertsLando:
LGTM

PR #369, opened 2020-04-22
add a unit test

chrisns:
Pushing just to get thoughts on tooling ([mocha](https://mochajs.org/), [sinon](https://sinonjs.org/), [chai](https://www.chaijs.com/)), code style (bdd, [should](https://www.chaijs.com/guide/styles/#should), formatting) etc.

In actual anger i'd like to refactor some of the code at the same time so that its more testable
for example for these tests move the `map`s out of the functions then i can stub them when testing a function with conditionals more easily.


chrisns:
I made a slack channel https://zwave2mqtt.slack.com/archives/C011XBWV7QX to discuss if thats easier

robertsLando:
Hi @chrisns, thanks for this, it is on my todo list from the start but I was always focused on features :rofl: I usually use other test suites like tap/tape but I think one or another it's the same the important is the result :) I think we could create tests for ZwaveClient and MqttClient separately and then some tests using Gateway and both clients. The problem is to test the zwave client we should create a FakeServer that emits events about nodes

chrisns:
yeah, i don't see the zwave (or mqtt) as a massive issue, it doesn't need to emit that much, just enough to prove that you've setup listeners.
cool, i'll plough on then, will make this not a draft PR when its a bit more meaningful

robertsLando:
@chrisns add you to contributors list :) https://github.com/OpenZWave/Zwave2Mqtt/commit/c30ad261d402ea7cfa255fb32c99fe00ba382cd5

robertsLando:
LGTM thanks @chrisns !

PR #373, opened 2020-04-23
linting - fix linter config - apply fixes to code - require linter to pass in CI

chrisns:


robertsLando:
@chrisns Do you use vscode? could you add also some vscode config? like auto fix linting on save?

chrisns:
yeah i use vscode
can do, depends what plugin(s) you've got setup to what config is required to wire it to eslint

a pre-commit hook using one of the many means of having that persisted that runs eslint and the tests is probably a more complete answer.

can we merge this and tackle that later? means I can get on refactoring code for tests without potential merge hell later :)

robertsLando:
Ok sure! LGTM

PR #372, opened 2020-04-23
add CT101 thermostat configuration

idyedov:
Created by copying CT32 configuration. Tested and it seems to work fine in home assistant

PR #375, opened 2020-04-23
reduce code duplication by putting the common 2gig thermostat definition in a variable/constant

chilicheech:
this is a follow up to: #372


robertsLando:
LGTM thanks @chilicheech :)

PR #390, opened 2020-04-24
test coverage lib/debug

chrisns:


robertsLando:
Essentially if DEBUG env is set I don’t automatically enable the debug module. Because without that instruction (debug.enable) you would not see any log in stdout/stderr, you should always specify the env var when starting z2m. 

We could think about move to pinojs logger for logging 


robertsLando:
The test so could just listen to stdout/err data event when that env var is set or not

chrisns:
Don’t want to test a third party module.
Just couldn’t spot on the debug obj what changed to be able to test it. Might just mock the external code with proxyquire

robertsLando:
@chrisns Nothing changes, the only thing that changes is the one I told you (stderr/stdout should or should not receive data).

robertsLando:
@chrisns Maybe this could help: https://github.com/visionmedia/debug/blob/master/src/common.js#L156

chrisns:
@robertsLando this is ready to review now if you have a mo, thanks

robertsLando:
Will look at it tomorrow, thanks Chris 🙏🏻 

---

Daniel - Software Engineer

Support me at:
    Github sponsors

Issue #357, opened 2020-04-16
[bug] Incorrect unit of measurement formatting with Hass Discovery for Zwave temperature sensors


raspberrypython:
v3.0.2

Build/Run method
- [x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

Zwave2Mqtt version: 3.0.2
Openzwave Version: 1.6.1061

**Describe the bug**
Home Assistant Hass Discovery is reporting temperature units in `C` or `F` rather than `°C` or `°F`.

The home assistant core code is defining `TEMP_CELSIUS` and `TEMP_FAHRENHEIT` as `°C` and `°F`:
https://github.com/home-assistant/core/blob/0.108.5/homeassistant/const.py#L353

Without the correct formatting, it is causing other components like HomeKit to fail with errors such as:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/src/homeassistant/homeassistant/components/homekit/type_sensors.py", line 96, in update_state
    temperature = temperature_to_homekit(temperature, unit)
  File "/usr/src/homeassistant/homeassistant/components/homekit/util.py", line 236, in temperature_to_homekit
    return round(temp_util.convert(temperature, unit, TEMP_CELSIUS), 1)
  File "/usr/src/homeassistant/homeassistant/util/temperature.py", line 29, in convert
    raise ValueError(UNIT_NOT_RECOGNIZED_TEMPLATE.format(from_unit, TEMPERATURE))
ValueError: C is not a recognized temperature unit.
2020-04-15 08:17:30 ERROR (MainThread) [homeassistant.core] Error doing job: Future exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/src/homeassistant/homeassistant/components/homekit/type_sensors.py", line 96, in update_state
    temperature = temperature_to_homekit(temperature, unit)
  File "/usr/src/homeassistant/homeassistant/components/homekit/util.py", line 236, in temperature_to_homekit
    return round(temp_util.convert(temperature, unit, TEMP_CELSIUS), 1)
  File "/usr/src/homeassistant/homeassistant/util/temperature.py", line 29, in convert
[...]

PR #394, opened 2020-04-25
map temperature units to hass values

idyedov:
https://github.com/OpenZWave/Zwave2Mqtt/issues/357

home assistant uses ```°C``` and ```°F``` for temperature units instead of plain ```C``` and ```F```.

Thus we need to either have zwave2mqtt convert it as part of the hass device JSON template, or handle it on home assistant side to override unit of measurement for every temperature sensor device.

This PR handles it on zwave2mqtt side.


robertsLando:
LGTM

PR #408, opened 2020-04-27
test coverage lib/jsonStore

chrisns:


chrisns:
@robertsLando I've increased the coverage checks to check all the code, so need you to override the check and force a merge into master and set our new baseline of accepted test coverage. (this will happen again when we do tests of the vue code)

package.json
```diff
@@ -46,8 +46,8 @@
     "start": "node bin/www",
     "lint": "eslint --ext .js,.vue .",
     "test": "mocha",
-    "coverage": "c8 --check-coverage --lines 100 --branches 100 --functions 100 mocha",
-    "record-coverage": "c8 report --reporter=text-lcov > ./coverage/lcov.info",
+    "coverage": "c8 --clean mocha",
+    "record-coverage": "c8 report --reporter=text-lcov --all --exclude=test/* --exclude=.git --exclude=.eslintrc.js --exclude=.postcssrc.js --exclude=wallaby.js > ./coverage/lcov.info",
```
robertsLando, line 50: Is there a way to put all this options in a file like `.c8`?

chrisns: nothing in their docs which is annoying, not read the source to see if its a hidden thing, could just put it in a `.c8` and then `$(cat .c8)` at runtime

robertsLando: keep that for another PR :)

PR #415, opened 2020-04-30
cache ci slow steps

chrisns:
brings builds from ~4mins down to <1m for routine stuff, can be tweaked to be a bit more refined if needed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: ci
 
-on: 
+on:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,47 +9,83 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+
     strategy:
       matrix:
         node-version: [8.x, 10.x, 12.x]
```
robertsLando, line 15: Should we add 14.x?

robertsLando:
Love this improvments ❤️

.github/workflows/ci.yml
chrisns: ... and drop `8`? ;)

robertsLando: Definitely :)

chrisns: in #416

PR #416, opened 2020-04-30
reflect what node supports https://nodejs.org/en/about/releases/

chrisns:


robertsLando:
@chrisns Just one question about cache. If openzwave master changes, is the cache updated?

chrisns:
> @chrisns Just one question about cache. If openzwave master changes, is the cache updated?

yes, it keys on the head commit sha https://github.com/OpenZWave/Zwave2Mqtt/blob/master/.github/workflows/ci.yml#L31

Issue #424, opened 2020-05-02
Display issue with labels overlapping input fields on mobile [bug]


djdd87:
**Version**

Build/Run method
- [ x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

Zwave2Mqtt version: 3.0.3
Openzwave Version: 1.6

**Describe the bug**
The new name and new location fields are overlapped by their labels on mobile (Pixel 3 Chrome, or emulated screensize on browser).

![image](https://user-images.githubusercontent.com/22151993/80862170-c9e6f580-8c6a-11ea-8cd0-7a85a43a68ab.png)

**To Reproduce**
1. Use a small browser window, or a mobile device (found on Pixel 3)
2. Click a z-wave node. 
3. See screenshot for issue.

**Expected behavior**
The label should not overlap the inputs as it makes it very hard to update on a mobile device.

**Additional context**
N/A

djdd87:
I've done a pull request but don't know how to link it to this issue.

PR #425, opened 2020-05-02
Bug #424 Fixing display issue on mobile

djdd87:
The labels overlapped the new name/location fields. Changed the xs sizing to 4 and 8 for the label and input respectively. Also made the ID full width.


chrisns:
Closes #424

PR #417, opened 2020-04-30
chore: major deps bump, code refactor and dark mode

robertsLando:

PR #432, opened 2020-05-04
respect the qos+retain config

chrisns:
some mqtt services don't support `retain` so use the shared config and also the shared `qos`


chrisns:
will merge after #417 , i think there'll be some conflicts with eslint around those lines

robertsLando:
@chrisns Move on with this

PR #447, opened 2020-05-06
docker build dispatch event

robertsLando:
I have found an interesting github action for events dispatch. Check this out. Zwave2Mqtt-docker repo has been already updated to support this, once this is merget it should trigger a new build there


robertsLando:
@chrisns Test must be green in order to push to master right now

robertsLando:
Let's see what happens now :laughing:

PR #458, opened 2020-05-07
fix: lgtm alerts

robertsLando:


app.js
```diff
@@ -1,26 +1,24 @@
...
-  cors = require('cors'),
-  ZWaveClient = reqlib('/lib/ZwaveClient'),
-  MqttClient = reqlib('/lib/MqttClient'),
-  Gateway = reqlib('/lib/Gateway'),
-  store = reqlib('config/store.js'),
-  config = reqlib('config/app.js'),
-  debug = reqlib('/lib/debug')('App'),
-  history = require('connect-history-api-fallback'),
-  utils = reqlib('/lib/utils.js');
+const express = require('express')
```
chrisns, line 1: switching to `const` is unfortunately going to cause pain when we write tests for this and we need to mock dependencies in to unit test. can we stick to `var` or `let` ?

lib/ZwaveClient.js
```diff
@@ -522,23 +520,22 @@ function sceneEvent (nodeid, sceneCode) {
...
   switch (notif) {
     case 0:
-      msg = 'node' + nodeid + ': message complete'
+    // msg = 'node' + nodeid + ': message complete'
       break
     case 1:
-      msg = 'node' + nodeid + ': timeout'
+    // msg = 'node' + nodeid + ': timeout'
       break
     case 2:
```
chrisns, line 531: this could probably be tidied up a bit?
like:
```js
case 0:
case 1:
case 2:
  break
```

lib/ZwaveClient.js
```diff
@@ -558,7 +555,7 @@ function notification (nodeid, notif, help) {
 
       break
     default:
```
chrisns, line 557: `default` seems redundant?

lib/ZwaveClient.js
robertsLando: sure

app.js
robertsLando: If it causes problems when writing tests I will use var instead, let doesn't have any sense as them are global

PR #467, opened 2020-05-11
feat: hass barrier_operator class support #445

### LINKED ISSUE #445: [bug] missing configuration values with hass integration
**Version**

Build/Run method
- [X] Docker
- [ ] PKG

Zwave2Mqtt version: 3.1.0 (on an rpi)
Openzwave Version: 

**Describe the bug**

I've got 3 devices which either don't configure the hass settings at all or only configure (seemingly) read-only data points for integration with hass.

Examples:
- minimote (in scene mode I believe they call it) creates 0 integration devices
- 2 thermostats (each adds 2 devices, air temperature (actual temp, not heat/cool desired temps), and battery level)
- garage door controller (4 read-only user values are created, but the 1 settable value (changing barrier state) is omitted)

**Expected behavior**

I would generally assume all 'user', 'configuration' and 'system' values create respective hass devices. Let me know if I'm missing something in how I've got it setup.

Thanks!

robertsLando:
Closes #445

### DIFF
--- a/hass/configurations.js
+++ b/hass/configurations.js
@@ -233,6 +233,24 @@ module.exports = {
     }
   },
 
+  barrier_state: {
+    type: 'cover',
+    object_id: 'barrier_state',
+    discovery_payload: {
+      command_topic: true,
+      state_topic: true,
+      value_template: '{{ value_json.value }}',
+      device_class: 'garage',
+      payload_open: 'Opened',
+      payload_close: 'Closed',
+      payload_stop: 'Stopped',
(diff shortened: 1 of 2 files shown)


robertsLando:
@chrisns Only coverage fails, could we merge this anyway?

robertsLando:
@chrisns sure but actually there is nothing that tests gateway/mqtt/zwave for now so that is why I’m ignoring coverage right now, the plan is to reach 100% for sure then each pr will have coverage as required check

PR #473, opened 2020-05-12
move docker build to this repo

chrisns:


.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,49 @@
...
+
+      - uses: crazy-max/ghaction-docker-buildx@v1
+        if: ${{ github.event_name != 'pull_request' }}
+
+      - name: Latest
+        if: ${{ github.ref == 'git/refs/heads/master'}}
+        run: |
+          docker buildx build \
+          --platform linux/386,linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x \
+          -t ${GITHUB_REPOSITORY}:latest \
```
robertsLando, line 38: Based on the old behaviour, `dev` tag should always be builded, `latest` and `<version>` tags are build just on releases

robertsLando:
I think we should use a single `buildx` step where `-t` options depends on the type of the trigger:

- Always build `dev` tag
- `latest` and `<version>` are builded just on release

So we could do something like 

bash
```
TAGS=`-t OpenZWave/Zwave2Mqtt:dev `

if [ "$GITHUB_EVENT_NAME" == "release" ]; then
 TAGS="$TAGS -t robertslando/zwave2mqtt:latest -t robertslando/zwave2mqtt:$VERSION" # make a new release
fi

docker buildx build \
          --platform linux/386,linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x \
         $TAGS \
```

.github/workflows/dockerbuild.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - uses: crazy-max/ghaction-docker-buildx@v1
         if: ${{ github.event_name != 'pull_request' }}
 
-      - name: Latest
+      - name: Dev
         if: ${{ github.ref == 'git/refs/heads/master'}}
```
robertsLando, line 34: Is this if needed as dev tag would always be builded?

chrisns: i'm relying on the build cache, so it'll fly through the second one. it's not elegant but docs on buildx aren't great :(

chrisns:
added `v` in the sed as github tags are prefixed with `v` and the docker hub

.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,58 @@
...
-         uses: peter-evans/repository-dispatch@v1
-         with:
-           token: ${{ secrets.DOCKER_REPO_TOKEN }}
-           repository: robertsLando/Zwave2Mqtt-docker
-           event-type: build
+      - uses: actions/checkout@v2
+
+      - name: Just build on PR
+        if: ${{ github.event_name == 'pull_request' }}
+        run: docker build .
```
robertsLando, line 20: Why this?

.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,58 @@
...
+          -t ${GITHUB_REPOSITORY}:latest \
+          --push \
+          .
+
+      - name: Latest
+        if: ${{ github.event_name == 'release' }}
+        run: |
+          docker buildx build \
+          --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 \
+          -t ${GITHUB_REPOSITORY}:latest \
```
robertsLando, line 47: You can use multiple `-t` options and merge this two steps. But I would prefer to merge all buildx xommands in a single step and set the tags like I explained in my comment [here](https://github.com/OpenZWave/Zwave2Mqtt/pull/473#issuecomment-627773095)

.github/workflows/dockerbuild.yml
chrisns: buildx was slow, figured it'd be nice to know the PR hasn't broken the ability to `docker build` for example changes to the dockerfile

chrisns: and didn't need to mutli-arch build to a reasonable degree of confidence on that

.github/workflows/dockerbuild.yml
chrisns: ah, didn't realise that worked and was what you meant, ok cool!

.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,50 @@
...
+          --push \
+          .
+
+      - name: Latest & tag
+        if: ${{ github.event_name == 'release' }}
+        run: |
+          docker buildx build \
+          --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 \
+          -t ${GITHUB_REPOSITORY}:$(echo ${GITHUB_REF} | sed "s/git\/refs\/tags\/v//") \
```
robertsLando, line 47: Are you sure that `GITHUB_REPOSITORY` will be lower case and not with upper cases? 

I mean: `OpenZWave/Zwave2Mqtt` is  not valid in dockerhub, it expects: `openzwave/zwave2mqtt`

.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,50 @@
...
+          docker buildx build \
+          --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 \
+          -t ${GITHUB_REPOSITORY}:latest \
+          --push \
+          .
+
+      - name: Latest & tag
+        if: ${{ github.event_name == 'release' }}
+        run: |
+          docker buildx build \
```
robertsLando, line 45: Do you know if, in case of a release, this command will build again all images or will be able to get them from cache from previous buildx command?

robertsLando:
@chrisns Nice work with kubernets docs too, thanks!

Just another thing, I would add also the old readme to that `docker` folder instead of having all the content in the main readme. What do you suggest?

.github/workflows/dockerbuild.yml
chrisns: it won't cache :(

.github/workflows/dockerbuild.yml
chrisns: hardcoded to avoid ambiguity

.github/workflows/dockerbuild.yml
robertsLando: So I suggest to use: 

```bash
TAGS="-t ${GITHUB_REPOSITORY}:dev"


if [ "$GITHUB_EVENT_NAME" == "release" ]; then
 VERSION=$(echo ${GITHUB_REF} | sed "s/git\/refs\/tags\/v//")
 TAGS="$TAGS -t ${GITHUB_REPOSITORY}:latest -t ${GITHUB_REPOSITORY}:${VERSION}" # make a new release
fi

docker buildx build \
         --platform linux/arm64/v8,linux/amd64,linux/arm/v6,linux/arm/v7,linux/386 \
         $TAGS \
```

chrisns: the issue is that it'll be triggered from different events, if you're quick _(~1hr based on current build time)_ at doing a release after a master merge maybe even concurrent, so a shell script with more logic wouldn't help here.
i guess if you waited for `master/dev` to build, the `latest`+`release` could be retagging the `:dev` tag.
so a cross arch pull, tag, push
buildx has awful docs at best, lacking features at worst :(

chrisns:
I did that initially, but there's so much in the original docker repo readme that is either not relevant, or docker documentation (like getting a shell in the container) that i filtered it out to merge.

.github/workflows/dockerbuild.yml
robertsLando: Based on this, what about create 2 different CI so? One that triggers on release (latest and version) and one on every master push (dev)? 

As you said, if for example we push a release to master followed by another push before the first one ends the second could prevent that the docker release its made?

chrisns: that is basically what happens, theres some `if` statements in there that play the right steps

it's not going to prevent anything, just be perceivably out of order if say the release finishes before the master branch does

chrisns:
Yes, if you’ve added the secrets?

robertsLando:
@chrisns humm seems it has skipped all build steps

chrisns:
I’ll have a look in a mo. Could you try a dummy release?

robertsLando:
I could create a beta release :)

.github/workflows/dockerbuild.yml
```diff
@@ -1,16 +1,50 @@
...
+      - name: Login to dockerhub
+        if: ${{ github.event_name != 'pull_request' }}
+        run: |
+          docker login --username $DOCKER_USERNAME --password $DOCKER_PASSWORD
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+
+      - uses: crazy-max/ghaction-docker-buildx@v1
+        if: ${{ github.event_name != 'pull_request' }}
```
robertsLando, line 31: that's strange this didn't worked

PR #474, opened 2020-05-12
apt-get update before installing deps

chrisns:
builds are currently failing


chrisns:
Yeah. I guess GitHub don’t maintain the apt cache that regularly

PR #493, opened 2020-05-18
fix docker build

chrisns:
fix what the expected value of `git.ref` should be


chrisns:
it should only do the `master` branch build, i'll figure out the release stuff in another repo, not sure whats going on there

robertsLando:
Seems working as now dev branch is compiling. Thanks @chrisns :)

PR #498, opened 2020-05-18
fix last references to docker image on openzwave/.

chrisns:


chrisns:
@robertsLando you missed a couple of references in #496

robertsLando:
I also just noticed that the `dockerbuild.yml` is actually building dev tag using latest tag :cry: could you fix that? Just replace `latest` with `dev` [here](https://github.com/OpenZWave/Zwave2Mqtt/blob/master/.github/workflows/dockerbuild.yml#L40)

PR #507, opened 2020-05-20
feat: support dynamically set min_temp and max_temp in climate #445

robertsLando:


lib/Gateway.js
```diff
@@ -753,6 +753,16 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
           setId = hassDevice.default_setpoint
         }
 
+        if (payload.max_temp && typeof payload.max_temp === 'string') {
```
chrisns, line 756: would be nicer (and more \*cough\* testable \*cough\*) to have this in a function? :)
```js
function myWizzBangFormatter(temp) {
  return temp && typeof temp === 'string' ? this.mqtt.getTopic(this.valueTopic(node, node.values[temp])) : temp
}

payload.min_temp = myWizzBangFormatter(payload.min_temp)
```

\*an `if` statement would probably be more readable than ternary

robertsLando: > and more *cough* testable *cough* 

:laughing: 

You right @chrisns, I need to edit my pr anyway as I misundestood what the user was telling me, he wants to take the value and not map the topic (that is not supported in hass)

chrisns: the code still looks basically the same, a function would be way better here

chrisns: when i say the same, both if statements and code are identical apart from variable names that could be parameterized into a function as i'd suggested in my first comment. if you do that i'll even write the test for you before merging ;)

robertsLando:
@chrisns What about

```js
isValueId (id) {
    return id ? id.split('-').map(v => parseInt(v)).filter(v => isNaN(v)).length === 0 : false
}
```

chrisns:
er confused!? that doesn't really make it more readable
I was more thinking along the lines of:

```js
function myWizzBangFormatter(in) {...}
payload.min_temp = myWizzBangFormatter(payload.min_temp)
payload.max_temp = myWizzBangFormatter(payload.max_temp)
```

chrisns:
where `myWizzBangFormatter` returning undefined is perfectly acceptable

robertsLando:
@chrisns I understand what you mean but return undefined is not like not having that field set, and Hass could reject the discovery payload. If I set that to undefined that property will still be visible in the hass payload

chrisns:
then do like
```js
myWizzBangFormatter(payload, 'min_temp')
```
?

lib/Gateway.js
```diff
@@ -466,6 +466,22 @@ function getMappedValuesTemplate (valueId, modeMap, integerList, defaultValue) {
...
+ * Retrives the value of a property from the node valueId
+ *
+ * @param {Object} payload discovery payload
+ * @param {String} prop property name
+ * @param {Object} node node object
+ */
+function setDiscoveryValue (payload, prop, node) {
+  if (typeof payload[prop] === 'string') {
+    const valueId = node.values[payload[prop]]
+    if (valueId && valueId.value != null) {
```
chrisns, line 479: `!==`

robertsLando: I have used `!=` null as it also works with `undefined`

chrisns: i'm still not sure about that, what other types are we anticipating it being?

robertsLando: What are you not sure about? Anyway we could make it !== and it should work anyway as I don't know when and why value could be undefined

robertsLando:
> i'm still not sure about that, what other types are we anticipating it being?

You mean the `valueid.value`? It could be a string/number

lib/Gateway.js
chrisns: you're relying on typecasting, i'm not sure I would grok it that way you're describing.

a more obvious (to me at least) way of expressing the same is:
```js
if (valueId && valueId.value)
```
though given we don't know the reprocrussions for other types maybe it would be better to check if its a string/number. if an object/array/etc snuck in there I imagine it'd break other things?

robertsLando: @chrisns What if `valueId.value` is 0? Your check would fail

chrisns: yes, i should have actually drank my ☕  before writing that!
ignore me, leave it with `!=`

PR #516, opened 2020-05-25
fix: payload parse of rgb dimmers #488

### LINKED ISSUE #488: [bug] Z2M scales values by 255 -> 99, for the first dimmer in a node
**Version**

Build/Run method
- [x] Docker
- [ ] PKG

Zwave2Mqtt version: `3.1.0` 
Openzwave Version: `1.6.1115`

**Describe the bug**
I was tracing why Z2M was changing my "set" value from 255 to 99. I'd prefer if Z2M did not scale my desired value, as I think this is what makes it hard for me to create proper integrations in HASS.

In [these lines](https://github.com/OpenZWave/Zwave2Mqtt/blob/f750ca6ec1d8a4955da3de0e493d41585a21033c/lib/Gateway.js#L485-L488), I saw that Z2M scales the value, only if the `object_id` of the device is `rgb_dimmer`. This gave me a way out - simply change the object id in the discovery doc, so Z2M doesn't hit that branch.

The bug here is that this only applies to the first dimmer. I have multiple nodes with multiple dimmers, where the object ids are named like this. I tested it, and surely enough, if I send a `255` to any of the `rgb_dimmer_N` devices, the value isn't scaled. This means that the first dimmer works very differently from all other dimmers.

image

**To Reproduce**
N/A

**Expected behavior**

* That Z2M does not scale my values in any way.
* Is it possible to avoid scaling - is changing the `object_id` a supported scenario?
  * I changed my `object_id`, but it didn't immediately have an effect - maybe I need to restart Z2M?
  * EDIT: I restarted Z2M, and now the scaling is completely gone. So another bug perhaps?

**Additional context**
N/A

robertsLando:
Fixes #488


robertsLando:
@chrisns ping

chrisns:
Could refactor into more functions to make it testable

robertsLando:
> Could refactor into more functions to make it testable

A dedicated function you mean?

PR #545, opened 2020-06-01
Update devices.js

jasonsinclair5:
Added Honeywell 39358 Fan Control so it will be discovered as a fan an not a light.

### DIFF
--- a/hass/devices.js
+++ b/hass/devices.js
@@ -183,6 +183,7 @@ module.exports = {
   '2-40976-266': [DANFOSS_TRV_ZWAVE], // Popp Radiator Thermostat
   '99-12340-18756': [FAN_DIMMER], // GE 1724 Dimmer
   '99-12593-18756': [FAN_DIMMER], // GE 1724 Dimmer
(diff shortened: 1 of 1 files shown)


hass/devices.js
```diff
@@ -183,6 +183,7 @@ module.exports = {
   '2-40976-266': [DANFOSS_TRV_ZWAVE], // Popp Radiator Thermostat
   '99-12340-18756': [FAN_DIMMER], // GE 1724 Dimmer
   '99-12593-18756': [FAN_DIMMER], // GE 1724 Dimmer
+  '57-12593-18756': [FAN_DIMMER], // Honeywell 39358 In-Wall Fan Control
```
chilicheech, line 186: please order this list alphanumerically. this entry should go above the 2 GE dimmers since the id starts with 57 which comes before 99

jasonsinclair5: Updated in the correct order.

robertsLando:
LGTM, thanks @jasonsinclair5

PR #549, opened 2020-06-03
Markdown linting

chrisns:

PR #550, opened 2020-06-03
prettier-standard

chrisns:
extending #549


robertsLando:
@chrisns I'm +1 for this, just a thought: what about add .vscode folder to the project too? With some settings like the defaultFormatter?

chrisns:
> @chrisns I'm +1 for this, just a thought: what about add .vscode folder to the project too? With some settings like the defaultFormatter?

you'll need to install the `numso.prettier-standard-vscode` plugin couldn't see an easy way to describe that in the settings.json :(

robertsLando:
> you'll need to install the numso.prettier-standard-vscode plugin couldn't see an easy way to describe that in the settings.json :(

Sincerly I don't know if there is a way but no problem you should get an info message suggesting you that plugin is missing

PR #540, opened 2020-05-30
Reverse proxy dynamic subpath

fabio-torchetti:
## What is it?

This is a PR to allow dynamic changing of the subpath the server is running behind. The idea is to allow the server to adapt to being behind a proxy that changes the root path.

## How?

The main `index.html` page is rendered on the fly taking into consideration the path the page is being served from. This can be determined in two ways:

  1 - A `X-External-Path` header is passed along with the request
  2 - A `base` is specified in the application configuration

  The api base and `socket.io` path is calculated based on the document `baseURI` is set in the `index.html` base tag.

## Why?

I am planning to have ZWave To MQTT run behind the `ingress` of Home Assistant. I really didn't like the option to expose it directly under a new domain. So, in the spirit of the Open Source _"If you don't like it, fix it!"_ this PR came to be.

This related to #317

PR #569, opened 2020-06-09
we need the views in the docker image now

chrisns:
#540 broke docker :(


robertsLando:
I’m from the phone and I can’t approve right now, feel free to merge this Chris, it’s ok :)

PR #574, opened 2020-06-10
Improve Support for Inovelli LWZ42

DelusionalAI:
The default generated hass device json doesn't include support for color temps, nor does it properly allow control over RGB. 
The format for color control of the bulb is #RRBBGGWWCW, but by default it tries to send #RRBBGG.

This change allows HA to color both RGB Color and White Color Temp. 

Comments, suggestions, or improvement are welcome.


DelusionalAI:
I’m not sure if the template would work with other light bulbs, but I don’t see a reason why it couldn’t. This is the only zwave bulb I have to test against. Maybe I can pick up a monoprice one though it’s one sale for $20... 


As long as the bulb expects RRGGBBWWCW I can’t see why it wouldn’t work. There’s nothing in here that’s unique to the invoelli one. 

If the bulb is just RGB and not RGBW, cutting out the color temp commands and cutting out the last 2 digits should work.

robertsLando:
@DelusionalAI The question is how do I know if it's rgb or RRGGBBWWCW? I should get the actual value from the buld and create a function like `isRgb` that sets a flag on the value

DelusionalAI:
I pulled it right from zwave2mqtt. 

RRGGBBWWCW

PR #576, opened 2020-06-11
chore: fix docker build and cross build even on pr

robertsLando:


robertsLando:
@chrisns Seems that `node:12.18.0-alpine` is missing 386 arch

```
[linux/386 internal] load metadata for docker.io/library/node:12.18.0-al...
#8 ERROR: no match for platform in manifest sha256:53cd04d947b6c097fde5ea031622f917e8a61378890effd8d548b1204628a3fc: not found
```

When removing that everything works

robertsLando:
Now `linux/386` fails in openzwave-shared build, seems something relative to the arch too. I suggest to remove `linux/386`

robertsLando:
Errors are still in the format of libopenzwave that comes from your container :confused:

robertsLando:
@chrisns I think the problem is the same about yesterday: https://github.com/OpenZWave/openzwave-docker/blob/master/Dockerfile.alpine#L1

You are using an sha here

PR #618, opened 2020-07-03
[devices] Add Intermatic PE653 MultiWave Receiver

j9brown:
This change depends on the following patch to OpenZWave: https://github.com/OpenZWave/open-zwave/pull/2289

### DIFF
--- a/hass/devices.js
+++ b/hass/devices.js
@@ -212,6 +212,101 @@ module.exports = {
       }
     }
   ],
+  '5-1619-20549': [
+    // Intermatic PE653 MultiWave Receiver
+    {
+      type: 'climate',
+      object_id: 'pool_thermostat',
+      values: ['49-1-1', '67-1-1'],
+      default_setpoint: '67-1-1',
+      discovery_payload: {
+        min_temp: 40,
+        max_temp: 104,
+        modes: ['heat'],
+        temperature_unit: 'F',
+        current_temperature_topic: '49-1-1',
+        current_temperature_template: '{{ value_json.value }}',
+        temperature_command_topic: true,
+        temperature_state_template: '{{ value_json.value }}'
+      }
+    },
+    {
+      type: 'climate',
+      object_id: 'spa_thermostat',
+      values: ['49-1-1', '67-1-7'],
+      default_setpoint: '67-1-7',
+      discovery_payload: {
+        min_temp: 40,
+        max_temp: 104,
+        modes: ['heat'],
+        temperature_unit: 'F',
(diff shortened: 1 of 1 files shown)


robertsLando:
Why did you added the switchs too? Aren’t them automatically discovered?

j9brown:
The switched ended up incorrectly labeled in confusing ways when auto
discovered.  It's a bit better with my patch to the open zwave devices
(removes a redundant switch instance) but I figured I would leave them in
here for completeness.

robertsLando:
Could you fix lint errors? Just run npm run lint-fix

j9brown:
Sure, I'll take care of it.

hass/devices.js
```diff
@@ -212,6 +212,101 @@ module.exports = {
       }
     }
   ],
+  '5-1619-20549': [
+    // Intermatic PE653 MultiWave Receiver
+    {
+      type: 'climate',
+      object_id: 'pool_thermostat',
```
robertsLando, line 219: Could you use more generic names? Is this device usually used to create this entities or is just for your use case?

hass/devices.js
```diff
@@ -212,6 +212,101 @@ module.exports = {
...
+        temperature_unit: 'F',
+        current_temperature_topic: '49-1-1',
+        current_temperature_template: '{{ value_json.value }}',
+        temperature_command_topic: true,
+        temperature_state_template: '{{ value_json.value }}'
+      }
+    },
+    {
+      type: 'climate',
+      object_id: 'spa_thermostat',
```
robertsLando, line 235: Same here. Then seems this are configured in the same way so you will have 2 thermostats that controls same values

robertsLando:
@j9brown News on this?

j9brown:
I’ll get back to you this afternoon.

hass/devices.js
j9brown: There are actually two thermostat instances, 67-1-1 and 67-1-7 but curiously only one water temperature sensor 49-1-1 that is used for both.  The thermostats have dedicated functions in the system.  One thermostat is used specifically for Pool temperature and the other for Spa temperature and they are labeled as such in the UI.

There's also support for air temperature and freeze sensors as well as some other auxiliary equipment I don't have, so that can be added by someone else later.  More information here if you're curious: https://www.intermatic.com/-/media/inriver/7092-8536.ashx

hass/devices.js
j9brown: These are the names used by the device itself, see next comment.

PR #639, opened 2020-07-23
fix: change default hass birth/will topic

### LINKED ISSUE #638: Home Assistant 0.113 change to birth and will message topic defaults
Per the Home Assistant 0.113 release notes, they have changed the MQTT integration to have a default topic for birth and will messages of "homeassistant/status". Details can be found in release notes at: https://www.home-assistant.io/blog/2020/07/22/release-113/

It would be ideal if Zwave2Mqtt either made the topic configurable, or if it would be updated to use the default value from HA. With HA going more and more towards UI based configuration, having to manually update the configuration YAML should ideally be an option, but not a requirement.

robertsLando:
Fixes #638

BREAKING CHANGE: Default Hass Birth/Will topic is changed from `hass/status` to `homeassistant/status` to reflect defaults of Hass 0.113

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,6 +408,9 @@ To enable this feature remember to set the flag **Hass Discovery** in Gateway se
 To achieve the best possible integration (including MQTT discovery):
 
 - In your **Zwave2Mqtt** gateway settings enable `Homeassistant discovery` flag and enable the MQTT **retain** too. The retain flag for MQTT is suggested to be sure that, once discovered, each device get the last value published (otherwise you have to wait for a value change)
+
+**NB:** Starting from version `4.0.0` the default Birth/Will topic is `homeassistant/status` in order to reflect defaults birth/will of Hass `0.113` th
+
 - In your **Home Assistant** `configuration.yaml`:
 
 ```yaml
@@ -416,10 +419,10 @@ mqtt:
   discovery_prefix: <your_discovery_prefix>
   broker: [YOUR MQTT BROKER] # Remove if you want to use builtin-in MQTT broker
   birth_message:
-    topic: 'hass/status'
+    topic: 'hass/status' # or homeassistant/status if z2m version >= 4.0.0
     payload: 'online'
   will_message:
-    topic: 'hass/status'
+    topic: 'hass/status' # or homeassistant/status if z2m version >= 4.0.0
     payload: 'offline'
 ```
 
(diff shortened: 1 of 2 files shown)


chrisns:
not read all the code, does this still use the topic prefix?

robertsLando:
> does this still use the topic prefix

Nope, it's a fixed topic :)

chrisns:
that seems like a bad thing, maybe my use case is unique but i run three zwave networks, one usa, one eu and another for another location. I don't run HA but if I did i'd probably want to separate the messages?

robertsLando:
@chrisns Nope, If you have 3 zwave network that are using the same broker and are publishing values all to the same hass instance it is correct, when the hass instance restarts all the 3 z2m instances will be notified about this. 

It's just a birth/will topic, it means that hass will publish there a message when it starts and when it stops.

chrisns:
ah ok, fine, but multiple HASS would be a problem then sharing the broker?

robertsLando:
I don't know why someone should do such thing but yes, in that case could be a problem (not really a big one, just when one has instance goes down even the other one will receive a rediscovery of the components, nothing that breaks anything it's just useless)

robertsLando:
@chrisns Waiting for that user asking for that feat :rofl: In that case I will make that option configurable

chrisns:
heh, i'd considered doing it when i pondered setting hass up :)

robertsLando:
Why don't use 2 different brokers?

robertsLando:
If you are interested I could make the support for this, just because it's you 😂

chrisns:
ha, hold fire until i try again to use HA

raetha:
As the original requester, to add my two cents. I'm now using the HA UI to configure my MQTT broker, in that scenario HA uses the default values suggested here. However, as far as I know you can still configure MQTT via YAML in HA and set alternate options for the birth and will messages. So though I'm only interested in the default, I could absolutely see others interested in having it configurable, since HA still allows that. But who knows for how much longer HA will allow it, as they slowly deprecate many things from YAML in favor of UI.

I also agree that having two HA instances point to one MQTT instance is likely to cause nothing but confusion or problems, but this being configurable would technically allow multiple Z2M and HA instances on the same MQTT broker.

robertsLando:
I absolutely agree @raetha, I just would like to keep configuration minimal to don't create any confusion, I have removed many other config in zwave for example because users don't like to read readme :confused: So I'm waiting for that user asking for this to see if it's really usefull or not

PR #659, opened 2020-08-03
create a security policy

chrisns:


robertsLando:
@chrisns Are emails correct? what is that `+zwave2mqtt`?

chrisns:
Just an alias
Means you can filter it

https://gizmodo.com/how-to-use-the-infinite-number-of-email-addresses-gmail-1609458192

robertsLando:
@chrisns Wow didn't know about that :open_mouth: LGTM

PR #671, opened 2020-08-17
feat: change meta theme-color based on theme #669

### LINKED ISSUE #669: feat: change theme-color header meta based on dark theme
**Is your feature request related to a problem? Please describe.**
This is related to when I select dark theme for the site on an Android phone and the navigation bar at the bottom remains white.

I would like the navigation bar to also go black or white according to whether dark theme is on or off. I believe this would mean setting the `<meta name="theme-color" content="#ffffff">` header to #000000 instead of #ffffff, but I'm not sure this is the fix. 

I also noticed that that meta tag is set twice here;
https://github.com/OpenZWave/Zwave2Mqtt/blob/master/views/index.ejs#L9
and here:
https://github.com/OpenZWave/Zwave2Mqtt/blob/master/views/index.ejs#L21
Might need to remove one of them.

**Describe alternatives you've considered**
Unfortunately, there are no alternatives that I can think of.

**Additional context**
Here's light mode with a light navigation bar at the bottom.
Screenshot_20200814-213605_Chrome

Here's dark mode with a light navigation bar at the bottom. This is the one I'd like the navigation bar to match the dark mode by being black.
Screenshot_20200814-213144_Chrome

Cheers,

P.S. apologies I don't know how to resize the images

robertsLando:
fixes #669

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -115,7 +115,9 @@ export default {
     },
     changeThemeColor: function () {
       var metaThemeColor = document.querySelector('meta[name=theme-color]')
-      var metaThemeColor2 = document.querySelector('meta[name=msapplication-TileColor]')
+      var metaThemeColor2 = document.querySelector(
+        'meta[name=msapplication-TileColor]'
+      )
 
       metaThemeColor.setAttribute('content', this.dark ? '#000' : '#fff')
       metaThemeColor2.setAttribute('content', this.dark ? '#000' : '#fff')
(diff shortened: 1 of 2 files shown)


robertsLando:
@chilicheech Could you give this a try?

chilicheech:
hi @robertsLando thanks for doing this. i tried testing this so i pulled the dev docker image and i'm still seeing the old index file. did it not get updated in the dev docker image yet?

robertsLando:
It should... let me check it tonorrow

PR #675, opened 2020-08-18
fix: improved watch logic of customDevices

### LINKED ISSUE #670: [bug] customDevices.json hot reload only works once
**Version** 4.0.0

Build/Run method
- [x] Docker
- [ ] PKG

Zwave2Mqtt version: 4.0.0
Openzwave Version: 1.6.

**Describe the bug**

vim (and probably other editors) update files by `rename`ing a new file on top of the old file. NodeJS's `fs.watch` watches based on inode, and so only watches the deposed "old file", after a single event that registers the renaming.

Instead, when `fs.watch` handler receives an eventtype `rename`, it needs to close the old watch and re-create the watch to watch the new file.

There's a potential race condition here wherein the file doesn't exist when it wants to be read, and coördinating with the desired "read at most once per sceond" behavior adds further complexity.

**To Reproduce**

start zwave2mqtt with a `customDevices.json` on disk. edit it with `vim`. observe that after the first time saving the file, zwave2mqtt no longer detects changes.

here's a minimal reproducer of the issue:

```nodejs
process.on('SIGINT', () => { console.log("exiting..."); process.exit(); })

const fs = require('fs')

const watch = (file) => {
  var lock = false
  fs.watch(file, () => {
    if(lock) { return }
    lock = true

    console.log("here, would read " + file)

    setTimeout(() => { lock = false }, 1000)
  })
}

console.log("waiting...")
watch("/work/test")
```

Run this like: `touch test && docker run --rm -it -v $PWD:/work node node /work/node.js`

Then play around with the `test` file in the same directory, edit it with vim, etc.

**Expected behavior**
zwave2mqtt should reload the file any time it's changed, regardless of whether it's deleted/re-created/renamed/etc. In fact, zwave2mqtt should load the file the first time it's created, even if it didn't exist before.

**Additional context**

pdbogen:
Although my specific use case is still broken because of the behavior of docker single file mounts, this appears to fix #670 

# Before

```
$ docker run --name=z2m --entrypoint=/bin/sh -it robertslando/zwave2mqtt:latest
$ echo '{}' > store/customDevices.json
$ node bin/ww
…
  z2m:Gateway Loaded 0 custom Hass devices configurations +0ms
…
```

Then, in a new window:
```
$ docker exec -it z2m /bin/sh
$ apk add vim
$ vim store/customDevices.json
…add something / change the file; `:wq`
```

Back in the first window, I see:
```
  z2m:Gateway Error while reading /usr/src/app/store/customDevices.json ENOENT: no such file or directory, open '/usr/src/app/store/customDevices.json' +20s
```

# After

```
  z2m:Gateway re-setting watch on /usr/src/app/store/customDevices.json +14s
  z2m:Gateway Error while reading /usr/src/app/store/customDevices.json ENOENT: no such file or directory, open '/usr/src/app/store/customDevices.json' (will retry) +1ms
  z2m:Gateway Loaded 1 custom Hass devices configurations +101ms
```

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -25,65 +25,74 @@ var allDevices = hassDevices // will contain customDevices + hassDevices
 
 debug.color = 2
 
-// try to watch a file for changes; if it disappears or doesn't exist watch the
-// directory, and retry the file watch if the directory shows changes
-const watch = (filename, fn) => {
-  var watcher = null
-  try {
-    watcher = fs.watch(filename, e => {
-      const exists = fs.existsSync(filename)
+// watchCustomDevices performs an initial loadCustomDevices, and then sets up a
+// watch on the file that was loaded. When that watch triggers, it calls
+// loadCustomDevices again. If loadCustomDevices ever loads a different file
+// than the initial file, the watch is switched to that new file.
+var customDeviceWatcher = null
+const watchCustomDevices = (last) => {
+  return () => {
+    var result = loadCustomDevices()
 
-      if (exists) {
-        fn()
-      }
+    if (result != last && customDeviceWatcher != null) {
+      // We'll be switching our watch.
+      customDeviceWatcher.close()
+      customDeviceWatcher = null
+    }
 
-      if (e === 'rename' || !exists) {
-        debug('re-setting watch on ' + filename)
-        watcher.close()
-        watch(filename, fn)
+    if (result) {
+      // Load succeeded!
+      if (customDeviceWatcher == null) {
+        // But we have no watcher, either because it was mis-matched or this
+        // is the first time; so set up a watcher.
+        customDeviceWatcher = fs.watch(result, watchCustomDevices(result))
       }
-    })
-  } catch {
-    debug(
-      'failed to set watch on',
-      filename,
-      ', will watch directory and try again'
-    )
-    watcher = fs.watch(path.dirname(filename), () => {
-      if (fs.existsSync(filename)) {
-        watcher.close()
-        watch(filename, fn)
-        fn()
-      }
-    })
+      return
+    }
+
+    // Load failed! If we need a watcher, set one on the directory
+    if (customDeviceWatcher == null) {
+      customDeviceWatcher = fs.watch(path.dirname(customDevicesJsPath), watchCustomDevices(null))
+    }
   }
 }
 
-var lastCustomDevicesLoad = 0
-// try to load the named file either as a .js lib if `lib` is true, or as JSON
-// otherwise
-function loadCustomDevices (filename, lib) {
-  if (Date.now() - lastCustomDevicesLoad < 1000) {
-    return
-  }
+const customDevicesJsPath = utils.joinPath(true, CUSTOM_DEVICES) + '.js'
+const customDevicesJsonPath = utils.joinPath(true, CUSTOM_DEVICES) + '.json'
 
-  var devices
+var lastCustomDevicesLoad = null
+// loadCustomDevices attempts to load a custom devices file, preferring `.js`
+// but falling back to `.json` only if a `.js` file does not exist. It returns
+// the path to the file that was loaded, or null, if a file could not be
+// loaded.
+const loadCustomDevices = () => {
+  var loaded = ""
+  var devices = null
+
   try {
-    if (lib) {
-      devices = reqlib(filename)
+    if (fs.existsSync(customDevicesJsPath)) {
+      loaded = customDevicesJsPath
+      devices = reqlib(CUSTOM_DEVICES)
+    } else if (fs.existsSync(customDevicesJsonPath)) {
+      loaded = customDevicesJsonPath
+      devices = JSON.parse(fs.readFileSync(loaded))
     } else {
-      devices = JSON.parse(fs.readFileSync(filename))
+      debug('no custom devices sources found')
+      return null
     }
-  } catch (error) {
-    if (error.message.indexOf('Cannot find module') >= 0) {
-      debug('Error loading', filename, ': not found')
-    } else {
-      debug('Error loading', filename, ':', error.message)
-    }
-    return
+  } catch(error) {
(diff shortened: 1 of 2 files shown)


lib/Gateway.js
```diff
@@ -25,17 +25,24 @@ var allDevices = hassDevices // will contain customDevices + hassDevices
...
-    fn()
+      fn()
+    })
+  } catch {
     setTimeout(() => {
-      lock = false
-    }, 1000)
-  })
+      watch(path, fn)
+    }, 100)
```
robertsLando, line 44: I would increese the timeout at least to 1/2 seconds

lib/Gateway.js
```diff
@@ -47,20 +54,36 @@ function loadCustomDevices (devices) {
   )
 }
 
+var lastCustomDevicesLoad = 0
+function tryLoadCustomDevices (file, first) {
+  if (Date.now() - lastCustomDevicesLoad < 1000) {
```
robertsLando, line 59: Is this to prevent the double call from watch?

lib/Gateway.js
```diff
@@ -47,20 +54,36 @@ function loadCustomDevices (devices) {
...
+    return
+  }
+
+  try {
+    const fileContent = fs.readFileSync(file)
+    loadCustomDevices(JSON.parse(fileContent))
+    lastCustomDevicesLoad = Date.now()
+  } catch (error) {
+    if (first) {
```
robertsLando, line 68: where is `first` var declaration?

lib/Gateway.js
pdbogen: This is to maintain the previous `lock` behavior. Personally I don't think the double call is that problematic, but I didn't want to change the behavior.

lib/Gateway.js
pdbogen: It's on line 58; it's an argument to `tryLoadCustomDevices`: https://github.com/OpenZWave/Zwave2Mqtt/pull/675/files/2a659f216c24c2d877e80d31623da12654f39f27#diff-cf3d598357eac50f80be79bac63d057aR58

lib/Gateway.js
robertsLando: I mean one or two seconds, not half second, sorry :)

lib/Gateway.js
robertsLando: Sorry missed it :)

lib/Gateway.js
pdbogen: ah, I think that's a little long. I'm sensitive to the fact that if the file is actually removed this will end up spinning pretty rapidly. 

But a few seconds means it's possible we'll actually miss some legitimate changes that happen quickly.

It'll add some complexity but perhaps I can limit re-tries instead, or implement back-off?

robertsLando:
@pdbogen Just one question. Let's think about a user that isn't using customdevices, that watch function will trigger every x seconds to check if one is created, is this safe or could it create leaks in a long running process? 

I ask this question also to @chrisns

pdbogen:
Ah, I didn't actually _intend_ to have it watch for the file even if it never existed; I accidentally removed that during refactoring. I am of the general opinion that it's better, because it can save on a process restart if a user wants to start using `customDevices.json`; but I'm happy to re-add the `existsSync` call on the initial `watch`.

(I was careful to consider the possibility of leaks. I don't _think_ it will create any, because no object is created in case `fs.watch` fails initially; and when we observe a `rename`, we take care to close the watcher.)

lib/Gateway.js
robertsLando: The main reason I would like an higher timeout is the one I have described in the following comment. As you said this is a pretty specific use case and I don't want that this could create any problems to users that are not using this. 

We could also enable this feature with an ENV var as an option or what do you mean with 'implement back-off'?
Also retry limit could be a valid alternative.

pdbogen:
I reviewed the `fs.watch` docs and played around a bit and realized I can watch the directory when the file doesn't exist, so instead of polling with `fs.watch` we can just wait. Any time something changes in the directory, we can try to watch the file again. Otherwise, nothing happens (at least on Linux, where this is implemented with `inotify`).

I also took the opportunity to clean up the `.js` load a bit and enable the same hot-reload functionality for it.

pdbogen:
Some aspects of the desired behavior here are unclear to me.

_I_ think that z2m should automatically load the files on the filesystem in the same way it would when starting up.

I think this is important, because if z2m _doesn't_ do this, then restarting z2m with no other changes except a weird customDevices state would cause a behavior change, and that seems harmful.

To support this, there are a few weird cases we have to handle:

1. `json` exists -- custom devices from json
2. `js` is created -- custom devices from js, because this is how z2m would load at start


1. neither file exists -- no custom devices
2. `js` is created -- custom devices from js
3. `json` is created -- custom devices from js (because we prefer js)
4. `js` is _deleted_ -- custom devices from json, because this is how z2m would load at start

pdbogen:
Some notes:

* we have to watch both files to get the behavior I describe. It's easier to watch `json` file than to only watch for it if `js` is deleted, as long as we always load correctly.
* I've refactored again the de-dupe logic, so that now we only perform the actual load if the content has changed. This is important and better than a timer, because this way we can be sure that we consider every possible event that could cause a change in load order.

robertsLando (approved):
Ok, seems legit. Thanks for this @pdbogen :) LGTM. I would like also @chrisns to take a quick look at this before merging

robertsLando:
@pdbogen I think tests are hanging because watcher doesn't allow the process to exit

robertsLando:
@pdbogen Could you fix tests?

pdbogen:
For sure, just haven't had the time to look yet. Would you prefer if the tests run without enabling the watches (probably easier), or if the tests somehow specifically terminate the watches before exiting (more like how the code will run in production)?

robertsLando:
@pdbogen The second one seems better :)

robertsLando (approved):
Perfect, thanks @pdbogen ! LGTM

Issue #720, opened 2020-09-12
[bug] Accented characters are not allowed for names and locations


besynnerlig:
**robertslando/zwave2mqtt:4.0.3**

**Describe the bug**
Previously accented characters were allowed to name devices and places in the control panel GUI. Now, there's a message telling me "Only a-zA-Z0-9_- chars are allowed". 

Example: "Tvättstugan" mening "Laundry Room"

I believe åäöüéÅÄÖÜÉ are allowed in MQTT messages node names so I see no reason to prevent entering them in the GUI.

robertsLando:
Just submitted a fix for this

PR #724, opened 2020-09-15
fix: extended allowed charaters in name/loc #720

robertsLando:


chrisns (approved):
Hate/love regex syntax

Issue #731, opened 2020-09-16
HASS CT32 thermostat C vs F [bug]


ve6rah:
**Version**

Zwave2Mqtt version: 3.2.2
Openzwave Version: 1.6.974

**Describe the bug**
This thermostat zwave communication works completely in Fahrenheit regardless of it's display setting, and regardless of the unit used in HASS. As a result, the MQTT parameter "temperature_unit": "F" must be set for this thermostat to work properly within a celsius setup, however by default that parameter is not set and must be configured manually.

**Additional context**
As this is an MQTT parameter that needs setting, I put this here, however if this is something that needs to be set in OpenZwave rather than in Zwave2Mqtt let me know and I'll move this bug.

chilicheech:
From what I understand the temperature unit is reported in `49-1-256` and `49-2-256` for the CT32, and the only value it supports is `Fahrenheit`.
![Screen Shot 2020-09-16 at 1 12 17 PM](https://user-images.githubusercontent.com/732659/93387545-42b3ac00-f81e-11ea-836b-44551ea29e87.png)
Here's my mqtt value for `49-1-256`
```
{
  "value_id": "31-49-1-256",
  "node_id": 31,
  "class_id": 49,
  "type": "list",
  "genre": "system",
  "instance": 1,
  "index": 256,
  "label": "Instance 1: Air Temperature Units",
  "units": "",
  "help": "Air Temperature Sensor Available Units",
  "read_only": false,
  "write_only": false,
  "min": 0,
  "max": 0,
  "is_polled": false,
  "values": [
    "Fahrenheit"
  ],
  "value": "Fahrenheit",
  "lastUpdate": 1599579139341
}
```
And it sets the `units` correctly for `49-1-1` as far as I can tell:
```
{
  "value_id": "31-49-1-1",
  "node_id": 31,
  "class_id": 49,
  "type": "decimal",
  "genre": "user",
  "instance": 1,
  "index": 1,
  "label": "Instance 1: Air Temperature",
  "units": "F",
  "help": "Air Temperature Sensor Value",
  "read_only": true,
  "write_only": false,
  "min": 0,
  "max": 0,
  "is_polled": false,
  "value": 76,
  "lastUpdate": 1600285031819
}
```
Where do you see a `temperature_unit` parameter in MQTT?

chilicheech:
Having said the above, this is what my hass discovery payload looks like:
```
{
  "min_temp": 50,
  "max_temp": 85,
[...]

PR #732, opened 2020-09-17
fix: set temperature_unit in climate discovery #731

robertsLando:
@chilicheech Check out this please

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -935,6 +935,10 @@ Gateway.prototype.discoverDevice = function (node, hassDevice) {
           payload.current_temperature_topic = this.mqtt.getTopic(
             this.valueTopic(node, currTemp)
           )
+
+          if (currTemp.units) {
+            payload.temperature_unit = currTemp.units.contains('C') ? 'C' : 'F'
+          }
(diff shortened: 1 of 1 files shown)


chilicheech:
Ah, yes, this looks good. I forgot that we have the `unit` field that we can use. That's why I was thinking of using `49-1-256`, but you're right, not every device has that.

Code looks good. I'm gonna test it on my setup. I'll do it as soon as I can. But don't wait for me to merge this.

chilicheech:
bah.. spoke too soon:
```
z2m     |   z2m:Gateway Error while discovering device climate_thermostat of node 31: currTemp.units.contains is not a function +0ms
```

chilicheech:
In my tests I replaced `contains` with `includes` and it worked:
I replaced `payload.temperature_unit = currTemp.units.contains('C') ? 'C' : 'F'`
with `payload.temperature_unit = currTemp.units.includes('C') ? 'C' : 'F'`

robertsLando:
Could you send a PR? Or I will do it tomorrow, not on my laptop right now :)

chilicheech:
Here's the PR #733 
https://github.com/OpenZWave/Zwave2Mqtt/pull/733
Cheers

robertsLando:
Thanks @chilicheech !

Issue #739, opened 2020-09-23
[question] How to handle unusual device values with home assistant discovery.


joshkay:
I'm trying to implement a control for my innovelli led bar. The switch is expecting the brightness value to be one of 0%, 10%, 20% etc. I cannot see a way to accomplish this using the [default MQTT light schema](https://www.home-assistant.io/integrations/light.mqtt/#default-schema).

Instead, I have managed to implement it using the [template MQTT light schema](https://www.home-assistant.io/integrations/light.mqtt/#template-schema).  I have this working using the template schema, but I cannot implement the RGB portion using that schema because it doesn't support RGB commands on a separate topic.

I have setup the custom device in customDevices.json.  Below is the config:
```
{
    "798-1-1": [
        {
            "type": "light",
            "object_id": "led",
            "values": ["112-1-13", "112-1-14"],
            "discovery_payload": {
                "schema": "template",
                "command_topic": "112-1-14",
                "command_on_template": "{{ (brightness / 255 * 10) | round(0) if brightness is defined else 5 }}{{0}}{{'%'}}",
                "command_off_template": "0%",
                "state_topic": "112-1-14",
                "state_template": "{{ 'off' if value_json.value == '0%' else 'on' }}",
                "brightness_template": "{{ value_json.value | regex_replace(find='%', replace='', ignorecase=False) | round(0) / 100 * 255 }}"
            }
        }
    ]
}
```

I'm wondering if there's a way to handle converting the brightness & rgb values on the zwave2mqtt side of things so it can accept a standard request from home assistant.

robertsLando:
@joshkay Have you tried to ask home assistant for a feature request? I mean, adding rgb commands on a separete topic?

joshkay:
I brought it up on Discord and they told me to use their implementation of OZW.  So I thought I'd see what your opinion was before pressing the issue with them.  Thoughts?

robertsLando:
How do you expect me to send that value? What's the format you need?

joshkay:
I was looking for a way to process the incoming/outgoing mqtt messages before sending the zwave command.

For example, if a brightness set command is sent and the value is 0-10, i would need to convert that to a string and add a % sign to the end of it before sending that command to the zwave device for it to respond properly.
Same thing with getting the value of the brightness, but in reverse.  Remove the % sign and send the number alone.

[...]

PR #741, opened 2020-09-24
feat: parse value functions #739

robertsLando:
This will allow users to specify a custom parse/recieve js function in gateway values table :)

Schermata da 2020-09-24 10-08-30

### DIFF
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-new-func */
 /* eslint-disable no-case-declarations */
 /* eslint-disable no-prototype-builtins */
 /* eslint-disable no-eval */
@@ -452,6 +451,7 @@ function evalFunction (code, valueId, value) {
   var result = null
 
   try {
+    /* eslint-disable no-new-func */
     var parseFunc = new Function('value', code)
     result = parseFunc(value)
   } catch (error) {
--- a/lib/Gateway.js
+++ b/lib/Gateway.js
@@ -247,7 +247,7 @@ function onValueChanged (valueId, node, changed) {
     }
 
     if (valueConf.parseSend) {
-      var parsedVal = evalFunction(valueConf.sendFunction)
+      var parsedVal = evalFunction(valueConf.sendFunction, valueId, tmpVal)
       if (parsedVal != null) {
         tmpVal = parsedVal
       }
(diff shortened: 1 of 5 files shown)


src/components/dialogs/DialogGatewayValue.vue
```diff
@@ -114,6 +114,52 @@
...
+                <v-switch
+                  label="Parse send"
+                  hint="Create a function that parse the value sent via MQTT"
+                  persistent-hint
+                  v-model="editedValue.parseSend"
+                ></v-switch>
+              </v-flex>
+
+              <v-container v-if="editedValue.parseSend">
+                <br />
```
chrisns, line 128: looks like you're using `<br />` to introduce some cosmetic spacing, can that be done with css, would that be better?

chrisns (changes requested):

lib/Gateway.js
```diff
@@ -430,6 +440,27 @@ function isValidOperation (op) {
...
+  var result = null
+
+  try {
+    var parseFunc = new Function('value', code)
+    result = parseFunc(value)
+  } catch (error) {
+    debug('Error eval function of value ', valueId.value_id, error.message)
+  }
+
+  return result
```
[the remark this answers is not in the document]
robertsLando: It works in this way, I would not return a promise as the conversion function should be fast. I could return the value yes but if I return the value by reading the code seems that the function has return that value. What do you think it's better?

src/components/dialogs/DialogGatewayValue.vue
robertsLando: I missed that. Removed

lib/Gateway.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-new-func */
```
chrisns, line 1: can we move these disable rules to the specific lines in question rather than the whole file where theres only a few use cases, otherwise eval's might creep in elsewhere for example

chrisns (changes requested):

lib/Gateway.js
robertsLando: about async functions, I really think there is no reason here. Users will only do some value parsing like transform an integer value to a rgb string and things like that, I think async will just make it more more complicated for no reason

chrisns: agree, just maybe make is clear in the gui that it must respond synchronously

robertsLando: done :)

chrisns (approved):

PR #746, opened 2020-09-28
feat(hass): Eurotronic Stella Z thermostat

markferry:
Create a climate device for Eurotronic StellaZ in home-assistant.

StellaZ functions the same as SpiritZ.

I tested:
 * temperature set point
 * current temperature
 * thermostat mode (Off, Heat, Cool)

The mode mapping is a bit weird since for Stella and Spirit it would normally be 'Heat' and 'Eco' rather than 'Heat' and 'Cool'. But perhaps there are reasons why Spirit was written that way.

### DIFF
--- a/hass/devices.js
+++ b/hass/devices.js
@@ -59,6 +59,29 @@ const THERMOSTAT_2GIG = {
   }
 }
 
+// Eurotronic Stella Z-Wave Thermostat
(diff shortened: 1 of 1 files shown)


robertsLando (approved):
LGTM. Thanks @markferry

PR #775, opened 2020-10-10
Misuse drying state for Full Power mode

peter-vanpoucke:
When you click the *boost* button on the *Eurotronic Spirit Z-Wave Plus Thermostat* it tends to set to the *Full Power* (Quick Heat) mode.

The details about that state mention:
> Full Power: Full power heating. This mode is left automatically after 5 minutes.

It would maybe be nice to use the fan mode (*auto* as standard and *high* during *Full Power*) for that, but I guess that'll need allot of tweaking.


robertsLando:
@peter-vanpoucke Could you run `npm run lint-fix` to fix lint errors?

peter-vanpoucke:
@robertsLando Done and also merged upstream.

Do you need any more detail about the change or should I look into the fan monde solution?

robertsLando:
@peter-vanpoucke I don't own this devices so it's hard for me to say it the discovery will work or not, I just try to check the correctness of the lines changed. Have you made any test locally? Is there something that isn't working?

robertsLando (approved):
From a code side LGTM. Let me know if I can merge this or you need to do more edits

peter-vanpoucke:
Ow clicked close by mistake.

No should be fine to merge.

robertsLando (approved):
lgtm

peter-vanpoucke:
Apparently I've use the wrong mode when putting this change. *drying* should've been *dry*. Will open an new pull for that.

PR #774, opened 2020-10-09
Update README.md

dbaker-rh:
Fix URL for kustomization.yaml

PR #781, opened 2020-10-12
chore(ci): fix build on PR

robertsLando:


robertsLando:
@chrisns I think the problem is node i386 is't supporting erbium anymore: https://hub.docker.com/r/i386/node/

Try to update to `dubmnium`

robertsLando:
Ok just saw your comment to #774 I will close this in favor of yours

PR #782, opened 2020-10-12
remove linux/386 build

chrisns:


robertsLando:
@chrisns I don't understand the reason to remove a tag from docker but ok :confused:

robertsLando (approved):
LGTM

PR #783, opened 2020-10-12
feat: improve HASS discovery of cover devices

### LINKED ISSUE #329: [bug]roller shutter detected as light dimmers
**Version**

Build/Run method
- [X] Docker
- [ ] PKG

Zwave2Mqtt version: v3.0.2
Openzwave Version: 1.6.1061

**Describe the bug**
all of my shutters are detected as light_dimmers, the function works but they are detected as light dimmers in homeassistant aswell.
i already have remove and readded the node since they where coming from homeassistant zwave integration that work in 1.4 openzwave.

any ideia how to fix the bug?
also, can i change the hass device json so at least they are detected as rollers in homeassistant?

{
  "type": "light",
  "object_id": "dimmer",
  "discovery_payload": {
    "schema": "template",
    "brightness_template": "{{ (value_json.value / 99 * 255) | round(0) }}",
    "state_topic": "zwave2mqtt/37/38/1/0",
    "state_template": "{{ \"off\" if value_json.value == 0 else \"on\" }}",
    "command_topic": "zwave2mqtt/37/38/1/0/set",
    "command_on_template": "{{ ((brightness / 255 * 99) | round(0)) if brightness is defined else 255 }}",
    "command_off_template": "0",
    "device": {
      "identifiers": [
        "zwave2mqtt_0xddaa6b5c_node37"
      ],
      "manufacturer": "FIBARO System",
      "model": "FGRM222 Roller Shutter Controller 2 (0x1000)",
      "name": "nodeID_37",
      "sw_version": "Unknown"
    },
    "name": "nodeID_37_dimmer",
    "unique_id": "zwave2mqtt_0xddaa6b5c_37-38-1-0"
  },
  "discoveryTopic": "light/nodeID_37/dimmer/config",
  "values": [
    "38-1-0"
  ],
  "persistent": false,
  "id": "light_dimmer"
}

p4p3r:
The main goal of this PR is to fix #329 

There are two main changes:
1. in the Zwave client, adding to the zwave nodes the basic, generic and specific device class as provided by the openzwave stack. I also added the constants for all the generic/specific classes.
2. in the Gateway, adapting the hass discovery logic to use the cover_position entity instead of the light_dimmer

I tested it with the Aeotec NanoShutter: https://github.com/OpenZWave/open-zwave/blob/master/config/aeotec/zw141.xml

This **only** solves discovery of the "cover position".
Looking at my Aeotec device I see many other controls that could be discovered in a more "specific" way. Unfortunately I don't know if this kind of device is more or less standardized or if any vendor comes up with different sets of commands and controls.
In my case, for example, to get the best behaviour I have to combine the cover_position entity with a "cover template" integration in Homeassistant.

Let me know if you want unit tests for these changes. I see there aren't many at the moment and I'm not sure what's your preferred approach.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -605,7 +605,7 @@ module.exports = {
     if (clsAttr) {
       return (
         clsAttr.specific[specificCls] ||
-        'unknownSpecificDeviceType' + specificCls
+        'unknownSpecificDeviceType_' + specificCls
       )
     } else {
       return 'unknownGenericDeviceType_' + genericCls
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -376,22 +376,22 @@ module.exports = {
     // https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave/const.py#L196
     // 0x00: specific_type_not_used // Available in all Generic types
     0x03: {
-      genericType: 'generic_type_av_control_point',
-      specificTypes: {
+      generic: 'generic_type_av_control_point',
+      specific: {
         0x12: 'specific_type_doorbell',
         0x04: 'specific_type_satellite_receiver',
         0x11: 'specific_type_satellite_receiver_v2'
       }
     },
     0x04: {
-      genericType: 'generic_type_display',
-      specificTypes: {
+      generic: 'generic_type_display',
+      specific: {
         0x01: 'specific_type_simple_display'
       }
     },
     0x40: {
-      genericType: 'generic_type_entry_control',
-      specificTypes: {
+      generic: 'generic_type_entry_control',
+      specific: {
         0x01: 'specific_type_door_lock',
         0x02: 'specific_type_advanced_door_lock',
         0x03: 'specific_type_secure_keypad_door_lock',
@@ -406,8 +406,8 @@ module.exports = {
       }
     },
     0x01: {
-      genericType: 'generic_type_generic_controller',
-      specificTypes: {
+      generic: 'generic_type_generic_controller',
+      specific: {
         0x01: 'specific_type_portable_controller',
         0x02: 'specific_type_portable_scene_controller',
         0x03: 'specific_type_portable_installer_tool',
@@ -416,43 +416,43 @@ module.exports = {
       }
     },
     0x31: {
-      genericType: 'generic_type_meter',
-      specificTypes: {
+      generic: 'generic_type_meter',
+      specific: {
         0x01: 'specific_type_simple_meter',
         0x02: 'specific_type_adv_energy_control',
         0x03: 'specific_type_whole_home_meter_simple'
       }
     },
     0x30: {
-      genericType: 'generic_type_meter_pulse',
-      specificTypes: {}
+      generic: 'generic_type_meter_pulse',
+      specific: {}
     },
     0xff: {
-      genericType: 'generic_type_non_interoperable',
-      specificTypes: {}
+      generic: 'generic_type_non_interoperable',
+      specific: {}
     },
     0x0f: {
-      genericType: 'generic_type_repeater_slave',
-      specificTypes: {
+      generic: 'generic_type_repeater_slave',
+      specific: {
         0x01: 'specific_type_repeater_slave',
         0x02: 'specific_type_virtual_node'
       }
     },
     0x17: {
-      genericType: 'generic_type_security_panel',
-      specificTypes: {
+      generic: 'generic_type_security_panel',
+      specific: {
         0x01: 'specific_type_zoned_security_panel'
       }
     },
     0x50: {
-      genericType: 'generic_type_semi_interoperable',
-      specificTypes: {
+      generic: 'generic_type_semi_interoperable',
+      specific: {
         0x01: 'specific_type_energy_production'
       }
     },
     0xa1: {
-      genericType: 'generic_type_sensor_alarm',
-      specificTypes: {
+      generic: 'generic_type_sensor_alarm',
+      specific: {
         0x05: 'specific_type_adv_zensor_net_alarm_sensor',
         0x10: 'specific_type_adv_zensor_net_smoke_sensor',
         0x01: 'specific_type_basic_routing_alarm_sensor',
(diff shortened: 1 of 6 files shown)


robertsLando:
@p4p3r  Really thank you for this PR! I Totally missed the methods `getNodeGeneric getNodeSpecific` in node OZW shared. I think this will open to lot improvments in hass discovery! 

In  order to fix the errors please remembere to run the command `npm run lint-fix`

lib/Constants.js
```diff
@@ -370,5 +370,241 @@ module.exports = {
...
+    },
+    0x30: {
+      genericType: 'generic_type_meter_pulse',
+      specificTypes: {},
+    },
+    0x0f: {
+      genericType: 'generic_type_non_interoperable',
+      specificTypes: {},
+    },
+    0x15: {
```
robertsLando, line 433: `0x15` is repeated twice in this map

lib/Constants.js
```diff
@@ -370,5 +370,241 @@ module.exports = {
...
+        0x01: 'specific_type_residential_hrv'
+      }
+    },
+    0x09: {
+      genericType: 'generic_type_windows_covering',
+      specificTypes: {
+        0x01: 'specific_type_simple_window_covering'
+      }
+    },
+    0x15: {
```
robertsLando, line 557: Dup

robertsLando (changes requested):

robertsLando:
> Let me know if you want unit tests for these changes. I see there aren't many at the moment and I'm not sure what's your preferred approach.

That would be much appreciated :) I don't create many unit tests because I'm focusing on bug fix and unfortunally I haven't much time for tests too

robertsLando:
> Unfortunately I don't know if this kind of device is more or less standardized or if any vendor comes up with different sets of commands and controls.

Based on my zwave experience I can tell you zwave should be standardized (in specs it is) but many devices have compatibility problems because them don't respect specs. If you identify some other discovery that could be improved with this feel free to add changes. 

What I can see it's maybe the `device_class` of the discovery payload could be better detected using this

p4p3r:
> @p4p3r Really thank you for this PR! I Totally missed the methods `getNodeGeneric getNodeSpecific` in node OZW shared. I think this will open to lot improvments in hass discovery!
> 
> In order to fix the errors please remembere to run the command `npm run lint-fix`

I'm fixing the linting issues now... and yes it actually caught a typo in the constants! 👍 
I'll push a new commit in a few minutes. I'm also running lint:prettier.

lib/Constants.js
p4p3r: Fixed in dd59a255189252f55124f938b40d0eb2cc52cfb1

lib/Constants.js
p4p3r: Fixed in dd59a255189252f55124f938b40d0eb2cc52cfb1

lib/Constants.js
```diff
@@ -370,5 +370,244 @@ module.exports = {
   },
   commandClass (cmd) {
     return this._commandClassMap[cmd] || 'unknownClass_' + cmd
+  },
+  _genericDeviceClassMap: {
+    // https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave/const.py#L196
+    // 0x00: specific_type_not_used // Available in all Generic types
```
robertsLando, line 376: Could you also add a reference to: https://github.com/OpenZWave/open-zwave/blob/master/config/device_classes.xml

p4p3r: Ah nice I imagined that this file existed somewhere in openzwave but for some reason I missed when looking for it! Thanks for pointing it out!

lib/Constants.js
```diff
@@ -370,5 +370,244 @@ module.exports = {
...
+    0x03: {
+      genericType: 'generic_type_av_control_point',
```
robertsLando, line 378: I think you could use just `generic` and `specific` as keys

lib/Constants.js
robertsLando: @p4p3r I just don't know how I could have missed it till now. From when I started developing hass discovery I always had problems to identify the device

chrisns (approved):

lib/Constants.js
p4p3r: Fixed in 7456e4d7b8dfc1197bcca9e247d2712019cc8265

lib/Constants.js
p4p3r: Fixed in ff15b8dc922b163dc5130109d38ad679ba1f4439

robertsLando (approved):
LGTM. Thanks @p4p3r !

robertsLando:
@p4p3r Let me know when you are done here

p4p3r:
@robertsLando I think now I'm done. My latest changes:
1. I added a couple of unit tests for the new functions I added in Constants.js
2. Double-checked the generic/specific device types with the xml file from openzwave. I fixed a few small things. I also sorted the keys as they are in the xml file (the unsorted version in homeassistant I initially used is probably more confusing and less maintainable)

I would love to add unit tests for `Gateway.discoverValue`, because as more device types are discovered there might be regressions, but I don't have time today :-(

robertsLando:
> I would love to add unit tests for Gateway.discoverValue, because as more device types are discovered there might be regressions, but I don't have time today :-(

That would be nice! Don't worry feel free to open another PR for that, merging this for now. Thanks! :heart:

robertsLando (approved):
LGTM

darkbasic:
Does this make the previous templates (https://github.com/OpenZWave/Zwave2Mqtt/pull/340) useless?

p4p3r:
> Does this make the previous templates (#340) useless?

Unfortunately not yet. This allow discovering cover_position entities instead of light_dimmer in a general way (no need for customDevices.js or special cases in devices.js)
AFAIK it's not possible yet to discover a cover template (I might go and look into that next). In my case I still have to combine the cover_position discovery with a specific template on HASS side (see my comment here: https://github.com/OpenZWave/Zwave2Mqtt/issues/329#issuecomment-707638886)

abmantis:
My Aeotec NanoShutters are still showing up as switch + light entities. Any tip how to troubleshoot it?

darkbasic:
The hassio maintainer bumped the zwave2mqtt version as per my request (https://github.com/hassio-addons/addon-zwave2mqtt/issues/64) but he didn't release a new version :(
I don't understand why he doesn't want to update it, but I think I'll drop the hassio addon for a plain docker installation.

robertsLando:
I suggest to use a plain docker install to be more free to use the preferred version

p4p3r:
A new release of the hassio addon has just been release and it includes zwave2mqtt 4.0.5 now.

darkbasic:
Too late, I already switched to plain Docker.

robertsLando:
@darkbasic, that's the best choice :)

PR #798, opened 2020-10-16
feat: Update last active on value changes

### LINKED ISSUE #490: [feat] Update 'Last active' when messages arrive from nodes
**Is your feature request related to a problem? Please describe.**
I understand [the `Last Active` column](https://user-images.githubusercontent.com/1027111/82088864-d18aad80-96f2-11ea-9830-dab3c531e5e1.png)) as the last time this specific node was active on my network. If a node was last active, days ago, it hasn't sent any message since then.

In reality, the column shows the last time a full refresh was done (?). This seems misleading.

That any time a message is received from a node (any kind of heartbeat, update, value, config, ..), the `Last Active` is amended to show this.

**Describe alternatives you've considered**
N/A

**Additional context**
I've identified [this code](https://github.com/OpenZWave/Zwave2Mqtt/blob/ac44aed7bf6cbdbf26609b583f4d4552bd0e089d/lib/ZwaveClient.js#L440) that seems to be the only place setting the `Last Active` value. 

Other functions in this file would need to have the same update. Perhaps a more generic place is available?

ahochsteger:
This updates the last active timestamp not only on status changes but on every value change a node reports.
I introduced a triggerStatusUpdate variable to prevent multiple updates within valueChanged().
Has been tested successfully on my local setup.

closes #490

### DIFF
--- a/lib/ZwaveClient.js
+++ b/lib/ZwaveClient.js
@@ -505,7 +505,6 @@ function nodeReady (nodeid, nodeinfo) {
 function valueChanged (nodeid, comclass, valueId) {
   var ozwnode = this.nodes[nodeid]
   var value_id = getValueID(valueId)
-  var triggerStatusUpdate = false
 
   parseValue(valueId)
 
@@ -522,20 +521,13 @@ function valueChanged (nodeid, comclass, valueId) {
     }
     // update cache
     ozwnode.values[value_id] = valueId
-
     // update last active timestamp
     ozwnode.lastActive = Date.now()
-    triggerStatusUpdate = true
   }
 
   // check if node is added as secure node
   if (comclass === 0x98 && valueId.index === 0) {
     ozwnode.secure = valueId.value
-    triggerStatusUpdate = true
-  }
-
-  // trigger a status update if something important has changed
-  if (triggerStatusUpdate) {
     this.emit('nodeStatus', ozwnode)
   }
 }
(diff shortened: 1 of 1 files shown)


lib/ZwaveClient.js
```diff
@@ -521,11 +522,20 @@ function valueChanged (nodeid, comclass, valueId) {
...
+
+    // update last active timestamp
+    ozwnode.lastActive = Date.now()
+    triggerStatusUpdate = true
   }
 
   // check if node is added as secure node
   if (comclass === 0x98 && valueId.index === 0) {
     ozwnode.secure = valueId.value
+    triggerStatusUpdate = true
```
robertsLando, line 534: This is useless

lib/ZwaveClient.js
```diff
@@ -521,11 +522,20 @@ function valueChanged (nodeid, comclass, valueId) {
...
+  }
+
+  // trigger a status update if something important has changed
+  if (triggerStatusUpdate) {
```
robertsLando, line 538: I'm not sure I want this, because this could cause some unexpected behaviour in the ui. If you have a selected node in the table every time a value of that node changes this is called and the ui will need to refresh all the tab content. IMO the triggerstatusupdate is useless and you could just leave there the lastactive update

robertsLando (changes requested):

lib/ZwaveClient.js
ahochsteger: Ok I'll remove the triggerStatusUpdate stuff.
So you mean that I shouldn't call `this.emit('nodeStatus', ozwnode)` at all after changing `ozwnode.lastActive`, right?

PR #582, opened 2020-06-12
[WIP] - Nodes table filtering

mrwho:
# Feature: enable filtering in nodes table

This PR enables filtering of the nodes by product name, device name, type, state and location.

It's  a _testing the waters_ PR, the functionality it's there, but I am not able to wrangle the filters into looking good.

## Code Changes

  Broke out the table into its own component, easier to understand what it's going on.

  Enable front-end testing with mocha and `esm` - now a dev dependency.

  I have also included the VSCode settings, we can cherry pick them out later. These configurations allow tests to run in VSCode for both client and server. 

  If you open the `vscode.code-workspace` you will see that it will create two main folders: `Server` and `Client`. This allows to specify different settings for the client and server code.

  If you install `hbenl.vscode-mocha-test-adapter` tests can be auto-run and debugged directly in the IDE for both backend and frontent.

PR #800, opened 2020-10-18
Add node table column filter

ahochsteger:
This PR adds the possibility to filter nodes by their column values which is helpful on a large number nodes.
Filtering can done by searching for a string in the column values or by selecting available values.

**Screenshots:**

Icons have been added to activate the filter options and to indicate a filtered column:
image

Example with a substring text filter:
image

Example with a selected value from a list of values:
image

It's already working very well but there are some things that may be improved, e.g.:
* Columns with type number and date may be filtered using a min and max filter (it is already prepared, but needs some more work)
* Layout of list selection may not work well with a large number of different values (check out wrapping and scrolling of items)
* Add a "Select all" or "Clear all" action for list items
* Optimize the code to be more efficient

### DIFF
--- a/src/components/ControlPanel.vue
+++ b/src/components/ControlPanel.vue
@@ -79,11 +79,62 @@
           :headers="headers"
           :items="tableNodes"
           :footer-props="{
-            itemsPerPageOptions: [10, 20, { text: 'All', value: -1 }]
+            itemsPerPageOptions: [10, 20, 50, 100, -1]
+          }"
+          :options="{
+            itemsPerPage: 100
           }"
           item-key="node_id"
           class="elevation-1"
         >
+          <template v-slot:header.type="{ header }">
+            <table-col-filter
+              :data="tableNodes"
+              :headers="headers"
+              :header="header"
+            ></table-col-filter>
+            {{ header.text }}
+          </template>
+          <template v-slot:header.product="{ header }">
+            <table-col-filter
+              :data="tableNodes"
+              :headers="headers"
+              :header="header"
+            ></table-col-filter>
+            {{ header.text }}
+          </template>
+          <template v-slot:header.name="{ header }">
+            <table-col-filter
+              :data="tableNodes"
+              :headers="headers"
+              :header="header"
+            ></table-col-filter>
+            {{ header.text }}
+          </template>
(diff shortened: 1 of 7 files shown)


src/components/ControlPanel.vue
```diff
@@ -903,6 +956,72 @@ export default {
...
+        fi.max !== undefined &&
+        fi.max !== null &&
+        typeof value === 'number' &&
+        value > parseInt(fi.max)
+      ) return false
+      if (!this.filterNodesByList(value, search, item, header.filterInfo.list)) return false
+      return true
+    },
+    filterNodesDateCol (value, search, item, header) {
+      return true
```
robertsLando, line 996: I think this is an error

robertsLando (changes requested):
Please also fix lint issues

robertsLando:
@ahochsteger Could you also check: https://github.com/OpenZWave/Zwave2Mqtt/pull/582

It is the same feature PR but it was also coming with unit tests

src/components/ControlPanel.vue
ahochsteger: Ah yes, that was the prepared code for extended date filtering - will remove it until it is done.

ahochsteger:
@robertsLando, thanks for pointing me to #582, I didn't know about it.
I will have a look and try to combine them as good as poosible - esp. the unit tests.

robertsLando:
Let me know when you want a new review of this

ahochsteger:
@robertsLando, @mrwho, after analyzing the PR #582 it seems to me the best approach would be to use #582 as the basis as since seems to be better structured and testable and apply the UI bits from this PR (#800) on top.
Would you agree?

What would be the best approach to deal with both PRs - create a new PR to mrwho:feature/table-filter, merge it on that side so it will become part of PR #582 and close PR #800 without merging?

robertsLando:
> t seems to me the best approach would be to use #582 as the basis as since seems to be better structured and testable and apply the UI bits from this PR (#800) on top.

@ahochsteger absolutely yes. My suggestion is to fork @mrwho branch, edit it with some bits of this (#800) and create a new PR. Feel free to close this

ahochsteger:
This PR is superseded by #823.

PR #817, opened 2020-11-04
Add `apt-get install -y git`

lucguinchard:


robertsLando:
@lucguinchard  Thanks for the pr but I think most users already have git installed (or at least know that it's required)

PR #823, opened 2020-11-09
feat: Add node table filter

ahochsteger:
This PR is a merged and improved version of #582 and #800 and adds the possibility to filter nodes by their column values.

Filtering can done by the following ways:
* Search for a sub-string
* Select from a list of values
* Set min/max number values
* Set min/max date values

**Screenshots:**

Icons have been added to activate the filter options and to indicate a filtered column:
image

Example of a substring and list selection filter:
image

Example of min/max number value filter:
image

Example of min/max date value filter:
image

### DIFF
--- a/src/components/nodes-table/filter-options.vue
+++ b/src/components/nodes-table/filter-options.vue
@@ -69,31 +69,33 @@
 export default {
   props: {
     value: {
-      type: Object,
+      type: Object
     },
     items: {
-      type: Array,
+      type: Array
     }
   },
   data () {
     return {
-      show: false,
+      show: false
     }
   },
   computed: {
     hasFilter () {
       return this.value !== undefined && this.value !== null &&
-        ((this.value.search !== undefined && this.value.search !== null && this.value.search !== '')
-        || (this.value.selections !== undefined && this.value.selections !== null && this.value.selections.length > 0)
-        || (this.value.min !== undefined && this.value.min !== null)
-        || (this.value.max !== undefined && this.value.max !== null))
-    },
+        (
+          (this.value.search !== undefined && this.value.search !== null && this.value.search !== '') ||
+          (this.value.selections !== undefined && this.value.selections !== null && this.value.selections.length > 0) ||
+          (this.value.min !== undefined && this.value.min !== null) ||
+          (this.value.max !== undefined && this.value.max !== null)
+        )
+    }
   },
   methods: {
-    showOptions() {
+    showOptions () {
       this.show = true
     },
-    hideOptions() {
+    hideOptions () {
       this.show = false
     },
     clearFilter () {
--- a/src/components/nodes-table/filter-options.vue
+++ b/src/components/nodes-table/filter-options.vue
@@ -1,11 +1,9 @@
 <template>
-  <v-menu
-    :value="show"
-    :close-on-content-click="false"
-    :offset-y="true"
-  >
+  <v-menu :value="show" :close-on-content-click="false" :offset-y="true">
     <template v-slot:activator="{ on, attrs }">
-      <v-icon small v-on:click="showOptions()"
+      <v-icon
+        small
+        v-on:click="showOptions()"
         v-bind="attrs"
         v-on="on"
         title="Filter options..."
@@ -16,7 +14,7 @@
     <v-card>
       <v-icon small v-on:click="hideOptions()" right>close</v-icon>
       <v-card-text>
-        <v-row v-if="value && value.type=='string'">
+        <v-row v-if="value && value.type == 'string'">
           <v-col>
             <v-text-field
               ref="search"
@@ -28,21 +26,43 @@
             ></v-text-field>
           </v-col>
         </v-row>
-        <v-row v-if="value && value.type=='number'">
+        <v-row v-if="value && value.type == 'number'">
           <v-col>
-            <v-text-field type="number" label="Min" v-model="value.min" clearable></v-text-field>
+            <v-text-field
+              type="number"
+              label="Min"
+              v-model="value.min"
+              clearable
+            ></v-text-field>
           </v-col>
           <v-col>
-            <v-text-field type="number" label="Max" v-model="value.max" clearable></v-text-field>
+            <v-text-field
+              type="number"
+              label="Max"
+              v-model="value.max"
+              clearable
+            ></v-text-field>
           </v-col>
         </v-row>
-        <v-row v-if="value && value.type=='date'">
+        <v-row v-if="value && value.type == 'date'">
           <v-col>
-            <v-text-field type="datetime-local" label="From" v-model="value.min" clearable></v-text-field>
-            <v-text-field type="datetime-local" label="Until" v-model="value.max" clearable></v-text-field>
+            <v-text-field
+              type="datetime-local"
+              label="From"
+              v-model="value.min"
+              clearable
+            ></v-text-field>
+            <v-text-field
+              type="datetime-local"
+              label="Until"
+              v-model="value.max"
+              clearable
+            ></v-text-field>
           </v-col>
         </v-row>
-        <v-row v-if="value && (value.type=='string' || value.type=='number')">
+        <v-row
+          v-if="value && (value.type == 'string' || value.type == 'number')"
+        >
           <v-col>
             <v-select
               v-model="value.selections"
@@ -56,9 +76,13 @@
             ></v-select>
           </v-col>
         </v-row>
-        <v-row v-if="value && value.type=='boolean'">
+        <v-row v-if="value && value.type == 'boolean'">
           <v-col>
-            <v-checkbox :indeterminate="value.bool == undefined || value.bool == null" v-model="value.bool" label="Boolean value"></v-checkbox>
+            <v-checkbox
+              :indeterminate="value.bool == undefined || value.bool == null"
+              v-model="value.bool"
+              label="Boolean value"
+            ></v-checkbox>
           </v-col>
         </v-row>
       </v-card-text>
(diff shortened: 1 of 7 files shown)


src/modules/NodeCollection.js
```diff
@@ -0,0 +1,77 @@
...
+  betweenNumber (properties, minValue, maxValue) {
+    return this.filter(properties, nodeValue =>
+      (minValue === undefined || minValue === null || minValue === '' || minValue <= nodeValue) &&
+      (maxValue === undefined || maxValue === null || maxValue === '' || maxValue >= nodeValue)
+    )
+  }
+
+  betweenDate (properties, minValue, maxValue) {
+    return this.filter(properties, nodeValue =>
+      (minValue === undefined || minValue === null || minValue === '' || new Date(minValue) <= nodeValue) &&
```
robertsLando, line 53: Isn't it better to use `.time()` and use timestamps to compare dates?

src/components/nodes-table/filter-options.vue
```diff
@@ -0,0 +1,109 @@
...
+            <v-text-field type="number" label="Max" v-model="value.max" clearable></v-text-field>
+          </v-col>
+        </v-row>
+        <v-row v-if="value && value.type=='date'">
+          <v-col>
+            <v-text-field type="datetime-local" label="From" v-model="value.min" clearable></v-text-field>
+            <v-text-field type="datetime-local" label="Until" v-model="value.max" clearable></v-text-field>
+          </v-col>
+        </v-row>
+        <v-row v-if="value && (value.type=='string' || value.type=='number' || value.type=='boolean')">
```
robertsLando, line 45: What about use `v-switch` for booleans?

src/modules/NodeCollection.js
ahochsteger: Yes, will change it.

robertsLando: Also I would create a method for this like `isUndefined(value) => return value === undefined || value === null || value === ''

src/components/nodes-table/filter-options.vue
ahochsteger: I'm not sure, if this will work here, since the secure property filter - while being a boolean - can have multiple states:
* Do not filter
* Filter for true values
* Filter for false values
* Filter for undefined values
That's why I kept v-select here.
Or do you have a better idea.

robertsLando: @ahochsteger Aren't states: `indeterminate` `true` and `false`? For filtering bools usually `checkboxes` are used and them can have an `indeterminate` state (that equals to undefined).

robertsLando: Check it here: https://vuetifyjs.com/en/components/checkboxes/#states

ahochsteger: Good idea - will change it to that.

src/modules/NodeCollection.js
ahochsteger: You are absolutely right, I also wanted to get rid of the repetitive and verbose undefined checking :-).

src/components/nodes-table/nodes-table.js
```diff
@@ -0,0 +1,117 @@
+'esversion: 6'
```
robertsLando, line 1: What is this for? Is it needed?

ahochsteger: I already removed this, not yet pushed.
It seems that there is some misconfiguration between VSCode and the npm build.
VSCode marks a lot of code to be invalid unless I add this directive - but it didn't help anyway.
Is there something I can do to configure VSCode to use the correct JavaScript language settings and code formatting? I didn't find anything in the README.

src/components/nodes-table/filter-options.vue
ahochsteger: A consequence of changing to 3-state v-switch would be that it would not possible anymore to just show nodes where the secure flag is undefined.
Do you still think, it's better to change it from v-select to v-switch?

src/components/nodes-table/nodes-table.js
robertsLando: Did you try to change `"editor.defaultFormatter": ` in .vscode/settings

ahochsteger: It's currently set to this:
`"editor.defaultFormatter": "numso.prettier-standard-vscode",`
What would be the correct value for Zwave2Mqtt?

src/components/nodes-table/filter-options.vue
robertsLando: Shouldn't all filter checks skip when switch is `indeterminate` (so the value `undefined` ) ?

src/modules/NodeCollection.js
ahochsteger: This is done.

src/components/nodes-table/nodes-table.js
```diff
@@ -0,0 +1,115 @@
...
+      return this.relevantNodes.values('status')
+    },
+    types () {
+      return this.relevantNodes.values('type')
+    },
+    lastActives () {
+      return this.relevantNodes.values('lastActive')
+    }
+  }
+}
```
robertsLando, line 115: In latest z2m version I store itemsPerPage on localstorage so when the user change the itemsPerPage it will remain the same after refreshing the page. I cannot see this here: f5ce7c2ea51337520d5512b6f944f9d35e36bcd2

src/components/nodes-table/nodes-table.js
robertsLando: @ahochsteger That should be the correct one but try with another if it shows errros

src/components/nodes-table/filter-options.vue
ahochsteger: You are right, that this would be inconsistent with other data types.
The only use case where filtering for undefined values might be for debugging Zwave2Mqtt itself but this may not be something a typical user would need.
If you think we don't need it, I will change it.

src/components/nodes-table/nodes-table.js
ahochsteger: Yes, I will include that.

src/modules/NodeCollection.js
```diff
@@ -0,0 +1,81 @@
...
+  betweenNumber (properties, minValue, maxValue) {
+    return this.filter(properties, nodeValue =>
+      (this._isUndefined(minValue) || minValue <= nodeValue) &&
+      (this._isUndefined(maxValue) || maxValue >= nodeValue)
+    )
+  }
+
+  betweenDate (properties, minValue, maxValue) {
+    return this.filter(properties, nodeValue =>
+      (this._isUndefined(minValue) || new Date(minValue).getTime() <= nodeValue) &&
```
robertsLando, line 57: Are you sure `nodeValue` is always a timestamp?

ahochsteger: No, I'm not sure, but it was while testing and debugging the code. Do you think it can have different data types?
Then I will wrap it again with new Date().

robertsLando: > Then I will wrap it again with new Date()

ahochsteger: Done.
I also moved `new Date(nodeValue)` into a `const nodeValueTime` to prevent creating the same object twice.

src/components/nodes-table/filter-options.vue
robertsLando: > If you think we don't need it, I will change it.

Ok!

robertsLando:
@ahochsteger Could you fix lint errors?

chrisns (approved):
really happy there's tests ❤️ 
I do wonder though if we've introduced a lot of alot of rigidity on the table, and how much harder it'll be to say add additional columns

src/components/nodes-table/nodes-table.js
ahochsteger: Ok, I found the setting in VSCode that caused many lines to be marked as warnings:
```JavaScript
"jshint.options": {
   "esversion": 6
}
```

ahochsteger:
> @ahochsteger Could you fix lint errors?

Sorry, but I don't find the reason for the following lint errors:
```
prettier-standard --check
Code style issues found in following files:
docker/store/nodes.json
docker/store/settings.json
src/components/nodes-table/filter-options.vue
src/components/nodes-table/index.vue
src/components/nodes-table/nodes-table.js
src/modules/NodeCollection.js
src/modules/NodeCollection.test.js
store/nodes.json
store/settings.json
```

It doesn't tell me what's wrong.
Do you have an idea?

chrisns:
@ahochsteger Run without the --check it'll do the fixes for you

src/components/nodes-table/nodes-table.js
robertsLando: @ahochsteger Did you include the commit?

robertsLando:
> I do wonder though if we've introduced a lot of rigidity on the table, and how much harder it'll be to say add additional columns

Not that harder, just a little bit

ahochsteger:
> @ahochsteger Run without the --check it'll do the fixes for you

Thanks, it's done.
That was easy :-)

ahochsteger:
> really happy there's tests 
> I do wonder though if we've introduced a lot of alot of rigidity on the table, and how much harder it'll be to say add additional columns

Kudos for tests go to @mrwho, who did the hard work in #582 on which this PR is based on.
But there seems to be a problem with the coverage/coveralls test that does not see see the new tests.
It reports 0% for NodeCollection.js which has many tests.

Concerning your comment on adding columns:
I tried to be as generic as possible for the table filter so it should be straight forward to extend for more columns.
You are right though, that you have to touch code on a few locations which may be able to encapsulate, but I guess it's out-of-scope for now and could be done, if more demand is here (e.g. if the feature to make it possible to dynamically choose the columns to be shown by the user).

src/components/nodes-table/filter-options.vue
ahochsteger: It's done, but I'm not really happy about the resulting layout spacing (caused by v-menu > v-card > v-card-text > v-row > v-col > v-checkbox):
image

I tried to reduce the nesting (at least for v-checkbox) down to v-menu > v-card > v-checkbox but that resulted in missing padding on the left that I was not able to add using CSS spacing helpers (e.g. pa-.. or ma-...) since they had no effect.
Being a first time Vuetify user I could need some help here.
If you have a better suggestion for the checkbox label, please let me know.

src/components/nodes-table/nodes-table.js
ahochsteger: @robertsLando Not yet, but I'm working on it.

ahochsteger: @robertsLando It's done.

chrisns:
> But there seems to be a problem with the coverage/coveralls test that does not see see the new tests.
> It reports 0% for NodeCollection.js which has many tests.

yeah, we need to fix the coverage collector, don't worry about that for now.
will pick it up as a separate exercise

robertsLando (approved):
LGTM. Thanks @ahochsteger ! :)

ahochsteger:
@robertsLando you are welcome - I'm looking forward to the next release :-).
I guess, #582 is now obsolete and may be closed.

PR #831, opened 2020-11-13
feat: Store settings for node table filter, sorting, show hidden nodes + settings refactoring

ahochsteger:
This feature makes sure that filters and sorting of the nodes table are persisted in the browser local storage on changes and survive a page refresh or browser restart.


src/components/nodes-table/nodes-table.js
```diff
@@ -37,6 +38,31 @@ export default {
...
+      if (
+        !filters ||
+        (Object.keys(filters).length === 0 && filters.constructor === Object)
+      ) {
+        filters = defaultValue
+      }
+      return filters
+    },
+    storeObject (key, val) {
+      localStorage.setItem(key, JSON.stringify(val))
```
robertsLando, line 64: you could make load and store more generic by checking the type of val like `typeof val === 'string' ? val : JSON.stringify(val)` so you can use the same functions for `node_itemsPerPage` local storage key

ahochsteger: Yes, I already thought about that :-).
I'll adjust the code and apply it to itemsPerPage too.

src/components/nodes-table/nodes-table.js
```diff
@@ -10,9 +10,10 @@ export default {
     filterOptions
   },
   data: () => ({
-    nodeTableItems: 10,
+    nodeTableItems: undefined,
```
ahochsteger, line 13: This is set to undefined for 2 reasons:
1. Default values should be set only once (see mounted())
2. This ensures that the default value is explicitely set and triggers the watch function to ensure there are always valid settings stored in the local store.

src/components/nodes-table/nodes-table.js
ahochsteger: @robertsLando Nevermind - the refactoring is already done and I now the settings 'nodes_showHidden' and 'dark' use the Settings module too.

robertsLando (approved):
LGTM. Thank you so much for this @ahochsteger :)

ahochsteger (commented):
I guess I hit the wrong button here - didn't want to start a code review, just had a question to the code I had to remove.

src/App.vue
```diff
@@ -219,9 +221,7 @@ export default {
       this.title = value.name || ''
     },
     dark (v) {
-      if (v) localStorage.setItem('dark', 'true')
-      else localStorage.removeItem('dark')
```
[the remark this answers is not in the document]
robertsLando: Nope don't worry, remove it it's like set it to false :)

robertsLando (approved):
LGTM

robertsLando:
@ahochsteger I'm not able to merge this, I think github has some problems today

robertsLando:
Schermata da 2020-11-16 10-09-37

@chrisns Any idea? I'm using `Squash and Merge` as always

robertsLando:
Hummm I think github has some bugs.... I press squash and merge and he has merged the PR :confused:

robertsLando:
@ahochsteger Don't worry this has happen also in some other repo I'm maintaining, I think github has some bugs today.

PR #842, opened 2020-11-29
Fix Table of Contents Links

bmight:
The broken table of contents links were driving me crazy so here's a PR to fix them. I couldn't tell if `README.md` is generated anywhere so I just updated it by hand.


robertsLando:
It's automatically generated and updated by a VS code plugin but seems that emojys in sections break it. Thanks for this fix

PR #844, opened 2020-12-02
Added Home Assistant support for HeatIt TRM3

hwestman:
Tested set point, mode and current temperature.

Not sure if you want to add the device id to the sample code. It wasn't clear to me, so I thought i'd suggest it.

### DIFF
--- a/hass/devices.js
+++ b/hass/devices.js
@@ -214,6 +214,27 @@ module.exports = {
       }
     }
   ],
+  '411-515-3': [
+    // Heatit Thermostat TF 058 (ThermoFloor AS)
+    {
+      type: 'climate',
+      object_id: 'thermostat',
+      values: ['67-1-1','64-1-0','49-1-1'],
+      mode_map: {'off': 'Off', 'heat': 'Heat'},
+      default_setpoint: '67-1-1',
+      discovery_payload: {
+          min_temp: 5,
+          max_temp: 40,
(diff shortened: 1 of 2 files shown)


hass/devices.js
```diff
@@ -214,6 +214,27 @@ module.exports = {
       }
     }
   ],
+  '411-515-3': [
+    // Heatit Thermostat TF 058 (ThermoFloor AS)
+    {
+      type: 'climate',
+      object_id: 'thermostat',
```
robertsLando, line 221: @hwestman Isn't it missing the `"setpoint_topic": { "Heat (Default)": "67-1-1", "Cool": "67-1-2" },`?

robertsLando (changes requested):

hass/devices.js
hwestman: "setpoint_topic: Key-Value object where keys are the modes of the Zwave thermostat and values are the matching setpoint value_id (use this if your thermostat has more than one setpoint)"

My conclusion was that it didn't have more than one setpoint, so i removed it. I'll look into it!

robertsLando: Sorry just noticed that you only have heat mode :) It's ok so

robertsLando (approved):
LGTM

robertsLando:
@hwestman Keep a look also to: https://github.com/zwave-js/zwavejs2mqtt

